PR #11, opened 2018-10-19
Add function to read ICGEM data files

### LINKED ISSUE #10: Function to read ICGEM data files

The [ICGEM Calculation Service](http://icgem.gfz-potsdam.de/) is an amazing resource to get gravity data. Their grids are in an ASCII format that can be tricky to parse because some values are included in the header. It would be great to have a function that reads the ICGEM file and returns an `xarray.Dataset`. 

This will really help with #9

santis19:
Port of the `load_icgem_gdf` function from `fatiando.datasets` to `harmonica`.
The new `load_icgem_gdf` loads gdf files that are downloaded from the
[ICGEM Calculation Service](http://icgem.gfz-potsdam.de/) with a few changes:

- Returns an `xarray.Dataset` instance instead of a dictionary.
- The `assert` lines have been replaced by proper `if...raise` checks.
- The `usecols` argument have been explicitly removed, but it still can be passed through `**kwargs`.
- The gdf file header is passed as a dictionary to the `attr` argument of `xarray.Dataset`.

Besides the new function, a few others changes have been made:
- Added `xarray` dependency to `environment.yml`, `requirements.txt`, `setup.py`, `doc/install.rst`.
- Added test functions for the new function.
- Added sample data file for tests functions.

Fixes #10 

**Reminders**


santis19:
@leouieda I've created the PR before the weekend. Is not ready to merge, I'll continue working next week.

leouieda:
:+1: no problem. Thanks for doing this!

If you add `WIP:` to your PR title the "WIP" status check will prevent merging.

santis19:
@leouieda 
I've make a few advances on this PR:

* I've written a few test functions for regular ICGEM gdf files and several for corrupted ICGEM gdf files.
I've made use of `tmpdir` of pytest in order to create temporary corrupted gdf files.
* The last line of the header is now read, i.e. the line containing the unit of each column (but it's only used to check file integrity).
* Some PEP8 related changes, mainly on the `with open(fname) as f:` lines: changed the `f` variable to a valid variable-name (`gdf_file`).

**There are a few issues I'd like to solve, but I think I need some help:**

1) `make test` shows missing coverage for lines [51-83](https://github.com/fatiando/harmonica/blob/d32d7b45d3c54d7ea3429edb063efff5c712cb6b/harmonica/io.py#L51-L83).
I don't understand how is pytest not testing those lines knowing that they are absolutely necessary in order to read the gdf file (and thus not failing the [first test function](https://github.com/fatiando/harmonica/blob/d32d7b45d3c54d7ea3429edb063efff5c712cb6b/harmonica/tests/test_icgem.py#L15)).

1) `make lint` shows a few errors on [`harmonica/io.py`](https://github.com/fatiando/harmonica/blob/d32d7b45d3c54d7ea3429edb063efff5c712cb6b/harmonica/io.py).
They are basically related with the fact that the `load_icgem_gdf()` function is too long (too many local variables, branches and statements).
Should we ignore these errors or we need to refactor the function by moving some lines to private functions?

leouieda:
@santis19 great work! The tests look really thorough! 

> make test shows missing coverage for lines 51-83.

Not sure. Sometimes this happens because not all `if` statements are executed. I'll have to look into it locally. The tests are failing on CI. I think you forgot to add the test files to `MANIFEST.in` and `setup.py`. 

> Should we ignore these errors or we need to refactor the function by moving some lines to private functions?

I put this check in the CI precisely so we wouldn't ignore them (which is my tendency as well). It's a good reminder that we should probably refactor this into smaller functions and test them separately. Try doing that and I think you'll find that the tests might be become simpler as well. Have the tests in mind when you refactor, i.e. try to refactor to make testing easier.

leouieda:
@santis19 I added the test data to `setup.py` so the CI tests should start passing now.

leouieda:
Coverage reports that have a `->` are indicating that certain branches of the code aren't being tested. For example, there is a test in which the `if` is true but there isn't one in which it isn't. I don't think this is an issue with this code so it's fine if some of the coverage is missing. But this is an indicator that it might not be the best implementation of this. I think we could have something better using a dict to match the conversions. This loop is a great target for becoming a separate function.

leouieda:
@santis19 great! I'll look at the code in a while.

The CIs are failing in 3.5 only so I don't think it's a problem with the CIs. Try making a 3.5 environment locally to test. I'm sure you'll get the error. I think this is related to `open` not looking the path object that you're giving it. I had similar problems in Pooch. Try coverting it to a string with `str` before passing to `open`.

harmonica/io.py
```diff
@@ -0,0 +1,157 @@
...
+    Barthelmes, F. and Kohler, W. (2016), International Centre for Global Earth
+    Models (ICGEM), in: Drewes, H., Kuglitsch, F., Adam, J. et al.,
+    The Geodesists Handbook 2016, Journal of Geodesy (2016), 90(10), pp 907-1205,
+    doi: 10.1007/s00190-016-0948-z
+
+    Parameters:
+
+    * fname: string
+        Name of the ICGEM .gdf file
+    * **kwargs:
```
leouieda:
```suggestion
**kwargs
```

harmonica/io.py
```diff
@@ -0,0 +1,157 @@
...
+        Arguments that will be passed to `numpy.loadtxt`.
```
leouieda:
```suggestion
        Extra keyword arguments to this function will be passed to :func:`numpy.loadtxt`.
```

harmonica/io.py
```diff
@@ -0,0 +1,157 @@
...
+
+    Returns:
```
leouieda:
```suggestion
    Returns
    -------
```

harmonica/io.py
```diff
@@ -0,0 +1,157 @@
...
+    * icgem_ds : xarray.Dataset
```
leouieda:
```suggestion
icgem_ds : :class:`xarray.Dataset`
```

harmonica/io.py
```diff
@@ -0,0 +1,157 @@
...
+        An `xarray.Dataset` with the data from the file.
```
leouieda:
```suggestion
        An :class:`xarray.Dataset` with the data from the file.
```

harmonica/io.py
```diff
@@ -0,0 +1,157 @@
...
+        The header of the gdf file is passed into the `attr` argument
```
leouieda:
```suggestion
        The header of the gdf file is passed into the ``attr`` argument
```

harmonica/io.py
```diff
@@ -0,0 +1,157 @@
...
+        of `xarray.Dataset`.
```
leouieda:
```suggestion
        of the :class:`xarray.Dataset`.
```

harmonica/tests/test_icgem.py
```diff
@@ -0,0 +1,286 @@
...
+    npt.assert_allclose(height, icgem_grd.height.values)
+    assert len(icgem_grd.data_vars) == 1
+
+
+def test_missing_shape(tmpdir):
+    "ICGEM file with missing shape"
+    fname = os.path.join(TEST_DATA_DIR, "icgem-sample.gdf")
+    attributes = ["latitude_parallels", "longitude_parallels"]
+    for attribute in attributes:
+        corrupt = tmpdir.join("missing_shape_" + attribute + ".gdf")
```
leouieda:
```suggestion
        corrupt = str(tmpdir.join("missing_shape_" + attribute + ".gdf"))
```

Turning the `tmpdir` into a string should be enough to satisfy Python 3.5. I think that `open` can only take non-string arguments starting in 3.6.

santis19:
@leouieda 
Thanks for the fixes in the docstring.
You were right about Python 3.5 failing tests. They also failed locally, but with the `str(tmpdir.join...)` fix they pass without any problem.
I've also added the `load_icgem_gdf` function to `doc/api/index.rst`.

I have a last question:
Why the tests are not raising PEP8 styling errors?, e.g. some lines are larger than 79 characters, but no test is failing.

Besides this minor doubt, I think that this PR is ready to be merged.
Maybe give it a final review and merge it if you agree.

leouieda:
> Why the tests are not raising PEP8 styling errors?

Yeah, that's a bit of a problem. Black doesn't format code in the docstrings and line-length isn't that strictly enforced anymore. I'll try to setup [pydocstyle](http://www.pydocstyle.org/en/3.0.0/) with the numpy style to get these. For now, try to build the docs locally and check that the pages you changed render correctly.

leouieda:
@santis19 made a few fixes to the doc style, please have a look. It's always good to check that pages render correctly and links work.

As soon as CIs pass, I'll merge this in. Thanks!

PR #17, opened 2018-10-31
Function to calculate Airy Isostasy

### LINKED ISSUE #15: Airy Isostasy calculations

We need a function that calculates the Moho depth from topography based on an Airy isostatic model. This will be useful for calculating isostatic gravity anomalies, for example.

The function only needs to know the topographic height values and crust, mantle, and water densities (optional):

```python
def isostasy_airy(topography, density_crust, density_mantle, density_water=None):
    ...
    return moho_depth
```

If water density is not given, negative topography will be ignored and the Moho depth for those values will be set to NaN.

The function can live in a new `isostasy.py` module.

I'm entirely happy with the function name. Any suggestions?

aguspesce:
Added `isostasy_airy` function to calculate the moho undulation assuming the Airy isostasy. 
The function calculates the moho undulation instead of moho depth because for the latter is necessary to assume a value for the normal crustal thickness. 
If water density is not given, bathymetry (negative topography) is ignored and the moho undulation for those values are set to NaN.

Fixes #15 

**Reminders**


aguspesce:
Hi @leouieda 
The function calculates the moho undulation because to calculate the moho depth is necessary to assume a value for the normal Earth. What is your idea ? depth or undulation?

For oceanic and continental crusts you assume the same density value, but it is possible to add different densities for each one. What do you think?

leouieda:
> What is your idea ? depth or undulation?

Good question. I think either is fine for me as long as it is very clear in the docstring which one we're calculating. So if you want to go with undulation, please add a definition for what that means in the docstring. Best to be extra clear in these matters because it's a common source of confusion and errors.

> For oceanic and continental crusts you assume the same density value, but it is possible to add different densities for each one. What do you think? 

Completely forgot about this, actually. It would be great if you could add that! :rocket: 

Thanks for taking the time to do this! Let me know if you need any help with the testing.

harmonica/isostasy.py
```diff
@@ -0,0 +1,54 @@
...
+    .. math ::
+        m = \frac{\rho_c - \rho_w}{\rho_m - \rho_c} h
+
+    where $L$ is the topography, $rho_m$ is the density of the mantle,
+    $rho_w$ is the density of the water, $\rho_c$ is the crustal density.
+
+    Parameters
+    ----------
+    density_mantle : float
+        Mantle density in kg/m$^3$.
```
leouieda: These `$` won't render into equations. Sphinx would require using `` :math:`kg/m^3` ``. Or you could use the ³ character, which is probably better in this case.

harmonica/isostasy.py
```diff
@@ -0,0 +1,54 @@
...
+
+    density_crust : float
+        Crustal density in kg/m$^3$.
+
```
leouieda: No need for these blank lines between the parameter descriptions.

harmonica/isostasy.py
```diff
@@ -0,0 +1,54 @@
+"""
+Function to calculate the moho undulation assuming the Airy isostatic hypothesis.
+"""
+import numpy as np
+
+
+def isostasy_airy(topography, density_crust, density_mantle, density_water=None):
+    """
+    Computes the moho undulation using the Airy hypothesis.
```
leouieda:
```suggestion
    Computes the Moho undulation from topography using the Airy hypothesis.
```

harmonica/isostasy.py
```diff
@@ -0,0 +1,69 @@
...
+
+    On oceanic points:
+
+    .. math ::
+        ar = \frac{\rho_{lc} - \rho_w}{\rho_m - \rho_{lc}} h
+
+        where :math:`h` is the topography/bathymetry, :math:`rho_m` is the
+        density of the mantle, :math:`rho_w` is the density of the water and
+        :math:`\rho_{uc}` and :math:`\rho_{lc}` are the density of the upper
+        and lower crust respectively.
```
leouieda: @aguspesce is there a reason for using the lower crust density in the oceans? Would it be possible to have a `density_oceanic` as well?

leouieda: Also, the identation of this block might cause trouble with the equation.

harmonica/isostasy.py
```diff
@@ -0,0 +1,69 @@
...
+
+def isostasy_airy(
+    topography,
+    density_upper_crust,
+    density_lower_crust,
+    density_mantle,
+    density_water=None,
+):
+    """
+    Computes the thickness of the roots/antiroots using the Airy's hypothesis.
```
leouieda:
```suggestion
    Computes the thickness of the roots/antiroots using Airy's hypothesis.
```

Nice! I like this wording a lot better. Avoids too many assumptions and is the way we see it written on text books.

leouieda:
@aguspesce the function looks really good! I merged in the latest changes from the master branch. Please run `git pull` locally to get these changes before making any new commits.

Should we get started on some tests? I think some test cases comparing results for very simple integer density and topography values (like 1, 2, 3) would be great start. Take a look at [`harmonica/tests/test_normal_gravity.py`](https://github.com/fatiando/harmonica/blob/master/harmonica/tests/test_normal_gravity.py) for some inspiration. Let me know if you need any help getting started.

Also, please leave a comment here when you want to me review your changes.

examples/isostasy_calculation.py
```diff
@@ -0,0 +1,54 @@
+"""
+Calculation of
```
leouieda: The example script needs to start with a title that will be used as the title for the plot in the gallery. For this one, `Airy Isostasy` should be good.

examples/isostasy_calculation.py
```diff
@@ -0,0 +1,54 @@
...
+density_oceanic_crust = 2900
+density_water = 1000
+
+# Root calculation considering the ocean
+root = hm.isostasy_airy(data.topography.values,
+                        density_upper_crust,
+                        density_lower_crust,
+                        density_mantle,
+                        density_oceanic_crust=density_oceanic_crust,
+                        density_water=density_water)
```
leouieda: This looks really good! You could probably set the densities directly in the function like:

```
root = hm.isostasy_airy(data.topography.values,
                        density_upper_crust=2670,
                        density_lower_crust=2800,
                        density_mantle=3300,
                        density_oceanic_crust=2900,
                        density_water=1000)
```

examples/isostasy_calculation.py
```diff
@@ -0,0 +1,54 @@
...
+data["root"] = (data.dims, root)
+
+# To obtain the depth of the Moho is necessary to assume a normal crust value
+T = 35000
+data["moho"] = - (T + data.root)
```
leouieda: I like that you put a note in there about needing the normal crust :+1:

examples/isostasy_calculation.py
```diff
@@ -0,0 +1,54 @@
...
+plt.colorbar(tmp, ax=ax, **cbar_kwargs).set_label("[meters]")
+ax.gridlines()
+ax.set_title("Isostatic root")
+ax = axes[1]
+tmp = data.moho.plot.pcolormesh(ax=ax, **pcolor_args)
+plt.colorbar(tmp, ax=ax, **cbar_kwargs).set_label("[meters]")
+ax.gridlines()
+ax.set_title("Isostatic depth Moho")
+plt.tight_layout()
+plt.show()
```
leouieda: Rather than plotting the root and Moho, it might be better to have one plot with the Airy Moho including the oceans and one without it (setting `density_water=None`) to show that this can be done.

examples/isostasy_calculation.py
```diff
@@ -0,0 +1,54 @@
+"""
+Calculation of
+"""
+import matplotlib.pyplot as plt
+import cartopy.crs as ccrs
+import harmonica as hm
+
+# Load the elevation model
+data = hm.datasets.fetch_topography_earth()
```
leouieda: It might be better to slice the data to get a single continent instead of doing the whole globe. Maybe Africa to avoid subduction zone? You can slice the `Dataset` using:

```
data_africa = data.sel(latitude=slice(lat_min, lat_max), longitude=slice(lon_min, lon_max))
```

examples/isostasy_calculation.py
leouieda: After, we just need a short paragraph about what you're showing in this example.

leouieda:
@aguspesce in the interest of time, I'm going to merge this as soon as the CIs are done. Great work! There are still some changes that I want to make but I'll open a new PR with them. I'll tag you so that you can see what I did and why.

Anyway, thanks for contributing this feature! It's really appreciated :1st_place_medal: :tada:

PR #22, opened 2018-11-07
Refactor normal gravity 

leouieda:
Made some corrections to the docstring and tweaks to the example plot
(put the colorbar at the bottom and use pcolormesh instead of contourf
for faster plotting).

While I was at it, I moved the calculations back into a single function.
It was actually really hard to read the code with the two auxiliary
functions. Plus, there was no way of testing these functions
individually so it made no difference having them separated. In the end,
it's better to have more readable code than pleasing pylint. I added an
exception to pylint regarding this function.

I also changed the module name to `gravity_corrections.py` so we can
include Bouguer etc in there as well. These are really small functions
so there is no need to have them in separate modules.

**Reminders**


leouieda:
> I'm a little bit confused about priorities: code readability, easier testing, good styling.

Ideally they are all priorities. This is the real challenge of software development. The coding itself is not that hard. But coming up with a good design that can be well tested and understood is really hard.

> These refactors give me an idea how to improve future PRs.

That's why I've been asking for you reviews :) Actually, @aguspesce might want to keep an eye on these things as well.

Also, please feel free to disagree with my design choices. They're not always good choices and I've come to regret many made in the original fatiando repo (or even in the recent Verde).

PR #23, opened 2018-11-07
Add global topography data and update gravity data

leouieda:
Use a 0.5 degree ETOPO1 dataset as Earth topography (will be used in
isostasy examples and Bouguer gravity). Update the global gravity data
to use EIGEN-6C4 at constant height to avoid dealing with conversion
between orthometric and geometric heights. The topography grid is in
geometric heights converted using an EIGEN-6C4 geoid.

Add functions to load these datasets using Pooch and examples for each.
Add an example calculating the gravity disturbance using the global
gravity data.

**Reminders**


data/examples/README.txt
```diff
@@ -0,0 +1,29 @@
...
+Github, the version is ``master``.
+
+You can change the base data directory by setting the ``HARMONICA_DATA_DIR`` environment
+variable to a different path.
+
+
+Available datasets
+------------------
+
+These are the datasets currently available:
```
santis19: Did you forget to add the datasets available?

leouieda: That README is what becomes the index page for sphinx gallery. So the examples are going to be rendered undetr that like so https://www.fatiando.org/verde/dev/sample_data/index.html

leouieda:
>  This PR is failing tests and docs building because the datasets cannot be found

I forgot to add a line to the CI builds to copy the sample data to the local folder. We don't the CIs to be making downloads every time. The downloading itself is well tested in Pooch so there shouldn't be a problem.

PR #24, opened 2018-11-08
Add sample airborne magnetic data from Rio

leouieda:
Use a subsection from the survey to limit file size. Original data are
made freely available by the Geological Survey of Brazil (CPRM) through
their [GEOSGB portal](http://geosgb.cprm.gov.br/).
Adds pandas as a dependency.

**Reminders**


leouieda:
Failures are due to #25. Should be good once that is merged.

PR #25, opened 2018-11-08
Use only strings in xarray.Dataset metadata

leouieda:
The `harmonica.load_icgem_gdf` function populated the `Dataset.attrs`
with metadata derived from the file header. Two of these fields
(`attributes` and `attributes_units`) were lists. This was fine while
using xarray with netcdf4 but it does not work if trying to use scipy to
save/load the `Dataset`. And since xarray 0.11 stopped installing
netcdf4 automatically, we can't load our sample gravity and topography
data without adding netcdf4 as a dependency.

Instead of doing that, `load_icgem_gdf` now converts the metadata to
strings. New copies of the two sample grids have been generated using
scipy (netcdf3).

cc @santis19 

**Reminders**


santis19:
I'll take this into account for future implementations.
In my opinion, we should try to minimize the dependencies, and sounds like `pydata` developers share this thought. And for our needs it does not substantially affect our functions and how we will define `Datasets`.

leouieda:
Yeah, this was a nasty surprise but it should be better now

PR #28, opened 2018-11-09
Add geodetic to geocentric coordinates conversion

### LINKED ISSUE #20: Function to transform geodetic to geocentric spherical coordinates

The tesseroid modeling is done in geocentric spherical coordinates (longitude, geocentric latitude, radius) but the data is usually in geodetic coordinates (longitude, geodetic latitude, geometric height). Proj4 apparently only converts the latitudes (https://proj4.org/operations/conversions/geoc.html) but not the height to radius. We need a function that does this conversion so that we can use tesseroids appropriately. 

The function will use the current ellipsoid (`get_ellipsoid`) to do the conversion. Tests can use simplified ellipsoids to check that the computations are correct (like setting the flattening to 1 and axis to an integer). It could go into a new `harmonica/coordinates.py` module and should look like:

```python
def geodetic_to_geocentric(latitude, height):
    ...
    return geocentric_latitude, radius
```

santis19:
Convert geocentric coordinates (defined by the ReferenceEllipsoid) to
geocentric (spherical) coordinates. Only geodetic `latitude` and `height`
are converted to `geocentric_latitude` and `radius`, geocentric and
geodetic latitudes are equal.  Add tests for new function: (1) test with
a "spherical" ellipsoid (in such case the geodetic and gecentric
coordinates are the same) (2) test with real ellipsoids on equator and
poles (on those points geodetic and geocentric latitudes are must be
equal).

Fixes #20

**Reminders**


santis19:
@leouieda 
The CI will fail because `ReferenceEllipsoid` does not have `first_eccentricty` among its attributes on this branch.
We should merge #27 before this.

leouieda:
:+1: Merged #27 so feel free to update/rebase this branch.

harmonica/coordinates.py
```diff
@@ -0,0 +1,47 @@
...
+Geographic coordinate conversion.
+"""
+import numpy as np
+
+from . import get_ellipsoid
+
+
+def geodetic_to_geocentric(latitude, height):
+    """
+    Convert from geodetic to geocentric coordinates.
```
leouieda:
```suggestion
    Convert from geodetic to geocentric spherical coordinates.
```

leouieda: Actually, this makes me think if we might want a different name for the functions. This might be confused with geocentric cartesian coordinates.

harmonica/coordinates.py
```diff
@@ -0,0 +1,47 @@
...
+    geocentric_latitude : float or array
+        The geocentric latitude (in degrees).
+    radius : float or array
+        The radial distance in the geocentric coordinate system (in meters), i.e.  the
+        distance between the point given by ``latitude`` and ``height`` and the origin
+        of the geocentric reference system.
+    """
+    ellipsoid = get_ellipsoid()
+    # Convert latitude to radians
+    latitude_rad = np.pi / 180 * latitude
```
leouieda:
```suggestion
    latitude_rad = np.radians(latitude)
```

harmonica/tests/test_coordinates.py
```diff
@@ -0,0 +1,56 @@
...
+
+def test_geodetic_geocentric_spherical_ellipsoid():
+    "Test geodetic to geocentric coordinates conversion if ellipsoid is a sphere."
+    rtol = 1e-10
+    sphere_radius = 1.0
+    # Define a "spherical" ellipsoid with radius equal to 1m.
+    # To do so, we define a zero flattening, thus an infinite inverse flattening.
+    spherical_ellipsoid = ReferenceEllipsoid(
+        "unit_sphere", sphere_radius, np.infty, 0, 0
+    )
```
leouieda: Nice! I'm increasingly happy with the ellipsoid mechanics in Harmonica.

harmonica/coordinates.py
leouieda: On the other hand, we also don't want a huge function name. Maybe just `geocentric_spherical`? Or maybe we leave it as is and trust that people will read the docs. But we might also need a geocentric Cartesian converter later on.

leouieda:
I merged in the changes from master so please remember to `git pull`

harmonica/coordinates.py
santis19: Every time I was writing the function name I needed to rethink if i wasn't confusing `geodetic_to_geocentric` with `geocentric_to_geodetic`, so sounds good idea to change its name.
Replacing geocentric with spherical makes it more rememberable and prevents the confusion with Cartesian geocentric coordinates, so I would propose `geodetic_to_spherical` in order to illustrate that the function will convert coordinates (I think that `geocentric_spherical` doesn't imply a coordinate conversion, could be confusing when reading/writing data processing code).

And keeping this naming rule, a future geodetic to Cartesian conversion could be named `geodetic_to_cartesian`.

leouieda: :+1: sounds good

leouieda:
@santis19 this looks good to me. 

Just thinking if we should standardize the API for things that take in coordinates. It might be good to always have `(longitude, latitude, height)` so that we make calls like `geodetic_to_spherical(*coordinates)` instead of `geodetic_to_spherical(coordinates[1], coordinates[2])`. What do think?

leouieda:
I'm going to merge this and we'll think about the arguments a bit more before making changes.

Thanks @santis19

PR #56, opened 2019-03-29
Forward modelling for tesseroids

### LINKED ISSUE #46: Tesseroid forward modeling

santisoler:
Start porting the forward modelling for tesseroids.
This implementation will use Numba for fast code execution.

- [x] Choose between 2D and 3D adaptive discretization
- [ ] Variable density in depth
- [x] Arbitrary GLQ degree on each direction
- [ ] Unwrap angles to prevent errors (|e - w| <= pi)
- [ ] Raise error when computation point is inside the tesseroid
- [ ] Raise error when computation point is on the surface of the tesseroid and 3D adaptive discretization

Fixes #46 

**Reminders**

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -22,6 +22,7 @@ Dependencies
 * `numpy <http://www.numpy.org/>`__
 * `scipy <https://docs.scipy.org/doc/scipy/reference/>`__
 * `pandas <http://pandas.pydata.org/>`__
+* `numba <https://numba.pydata.org/>`__
 * `xarray <https://xarray.pydata.org/>`__
 * `attrs <https://www.attrs.org/>`__
 * `pooch <http://www.fatiando.org/pooch/>`__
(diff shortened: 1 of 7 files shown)


santisoler:
@leouieda Would you like to take a look at this? I like how this Tesseroid implementation is looking, but want to have other opinions.

So far I've implemented the adaptive discretization (2D and 3D) and the conversion from Tesseroids to point masses. A lot is left to be done, but I think we are on track.

santisoler:
The implementation of this enhancement has been moved to #60

PR #60, opened 2019-04-29
Forward modelling for tesseroids and point masses

### LINKED ISSUE #46: Tesseroid forward modeling

### LINKED ISSUE #47: Point masses gravity computation on geographic coordinates

santisoler:
Add functions for computing gravitational fields generated by tesseroids and point
masses in geocentric spherical coordinates.
Forward model for tesseroids only admit constant densities.
Implemented 2D and 3D adaptive discretization and editable GLQ orders.
Tesseroids are converted to point masses following GLQ and then make use of point mass
forward models to compute the gravitational fields.
Add two pytest runs: one with Numba jit disabled to check coverage, and another one with
Numba jit enabled.
Add a decorator for test functions that must be run only if Numba jit is enabled.
Add a simple example on forward modelling with tesseroids.

Both #56 and #54 were merged into a single PR as proposed by @leouieda.
Because both codes will be intimately related, it's better to discuss them 
simultaneously on the same PR.

Fix #46 fix #47 

**Reminders**


santisoler:
The new idea is to build low-level kernel functions for the tesseroid and point mass forward modelling.
Because a single high-level forward modelling function cannot be as general as the many possible uses of it (e.g. computation of a single element for testing or computation performance, forward modelling of a mesh or a relief, inversion method that uses geometric symmetries to improve performance, etc), we decided to implement several public functions, one for each need, but they will all ultimately make use of the same low-level kernel functions.

harmonica/forward/tesseroid.py
```diff
@@ -0,0 +1,228 @@
...
+        coordinates, tesseroid, distance_size_ratio
+    )
+    # Initialize result
+    result = 0
+    jit_point_mass_gravity(
+        longitude, latitude, radius, point_mass, kernels[field], result
+    )
+
+
+def tesseroids_to_point_masses(tesseroids, glq_degrees=GLQ_DEGREES):
```
leouieda: This function needs to be  jit-compiled. The numpy functions that allocate memory should be moved out of it. Allocation is slow so having the arrays as inputs is better.

santisoler:
@leouieda According to Numba's documentation [list comprehension are supported](https://numba.pydata.org/numba-doc/latest/reference/pysupported.html#list-comprehension).
Numba complains if I pop the tesseroid from the stack like this:
```python
tesseroid = stack[stack_top, :]
```
But there's no problem (and according to the documentation it should run fast) if I use list comprehension:
```python
tesseroid = [stack[stack_top, i] for i in range(6)]
```

I've implemented the last code on [L148](https://github.com/fatiando/harmonica/pull/60/commits/468e77defdbff727c1b9ee53f5dc7a311ceb7ff7#diff-9bda0378cd7c9c474f82577a0b5bdbd1R148).

leouieda:
@santisoler what if you pop it as `stack[stack_top]`?

santisoler:
> @santisoler what if you pop it as `stack[stack_top]`?


Numba complains because I'm passing a `tesseroid` as a `list` and not a `numpy.array`:
```
Cannot unify reflected list(float64) and array(float64, 1d, C) for 'tesseroid', defined at harmonica/forward/tesseroid.py (161)
```

Although `tesseroid = [stack[stack_top, i] for i in range(6)]` has the same problem if the `tesseroid` is a `numpy.array` instead of a `list`.

Thanks for making me test this!

So we should choose one of them and convert the `tesseroid` to the right data structure for the corresponding code.
Because we want to work with `numpy.array`s inside the jitted functions, I would go for your new alternative and transform the tesseroid into a `numpy.array` on the `tesseroid_gravity` function. Do you agree?

santisoler:
I've made some of the changes we discussed today (besides the ones to the adaptive discretization).
Could you take a quick look and see if there's anything we should change?

PS: I haven't added the proper forward modelling, i.e. the computation of the gravity field, because I'm thinking to add a point mass function that takes several point masses and a single computation point.
I know you said that previous Fatiando implementation used Python for loops to run over the tesseroids, but the computation of the gravity fields of the GLQ nodes were performed under Cython/Numba.

leouieda:
>  I would go for your new alternative and transform the tesseroid into a numpy.array on the tesseroid_gravity function. Do you agree?

Yep, that's best. We want to avoid lists and other Python types as much as possible.

> Could you take a quick look and see if there's anything we should change?

:+1:

santisoler:
@leouieda Now both tesseroid and coordinates are converted to numpy arrays.
From now on I'll write jitted functions considering that arguments **should** be arrays and not lists.
Considering that requirement from the start will surely prevent some bugs in the future and improve our design.

harmonica/forward/point_mass.py
```diff
@@ -0,0 +1,292 @@
...
+    """
+    Compute gravitational fields of a point mass on spherical coordinates.
+
+    Parameters
+    ----------
+    coordinates : list or array
+        List or array containing `longitude`, `latitude` and `radius` of computation
+        points defined on a spherical geocentric coordinate system.
+        Both `longitude` and `latitude` should be in degrees and `radius` in meters.
+    point_mass : list or array
```
leouieda: Maybe we could use just `point` as it sounds weird reading `point_mass, mass`.

harmonica/forward/point_mass.py
```diff
@@ -0,0 +1,292 @@
...
+    mass : float
+        Mass of the point mass in kg.
+    field : str
+        Gravitational field that wants to be computed.
+        The available fields are:
+
+        - Gravitational potential: ``potential``
+        - Accelerations or gradient components: ``gx``, ``gy``, ``gz``
+        - Maurssi tensor components: ``gxx``, ``gxy``, ``gxz``, ``gyy``, ``gyz``,
+          ``gzz``
```
leouieda: It's best to put `(gravity gradient)` since not everyone uses the Marussi name.

leouieda: We should also move away from `x, y, z` in the names. Maybe change to `n, e, r` in this case. That's more straight forward.

leouieda:
@santisoler maybe it would be best to focus on implementing and testing only the g_radial instead of everything. This is a lot of work and there is nothing too different about each component anyway. They can be added in later PRs so we can keep this one from growing too large and taking too long.

harmonica/forward/point_mass.py
santisoler: :+1: you're right. The idea that we are computing the gravitational field from a point mass is very clear on the function name. So we can name its coordinates just `point`.

harmonica/forward/point_mass.py
santisoler: See the comment bellow.

santisoler:
> @santisoler maybe it would be best to focus on implementing and testing only the g_radial instead of everything. This is a lot of work and there is nothing too different about each component anyway. They can be added in later PRs so we can keep this one from growing too large and taking too long.

Yes, I agree. This way we can focus more on the design rather than making this PR too cluttered with the full set of kernels.
So, I'll remove the kernels we don't want to implement for now, what leaves your latest comment obsolete (the one that proposes to change the name of `x`, `y` to `n`, `e`).
I'll just rename the `g_z` to `g_radial` as you propose.

harmonica/forward/tesseroid.py
```diff
@@ -0,0 +1,280 @@
...
+    # Apply adaptive discretization on tesseroid
+    n_splits, error = _adaptive_discretization(
+        coordinates,
+        tesseroid,
+        distance_size_ratio,
+        stack,
+        small_tesseroids,
+        radial_discretization=three_dimensional_adaptive_discretization,
+    )
+    if error == -1:
```
santisoler: @leouieda I think we should move this error rising to `_adaptive_discretization`.
Firstly because Numba supports Exceptions raising: [Numba Supported Python features](https://numba.pydata.org/numba-doc/dev/reference/pysupported.html#constructs).
Secondly, if we leave the error handling on the high-level functions, we will have to copy-paste these lines on every high-level function. That will be annoying and will have to make test functions for each time it appears.

What do you think?

santisoler: I've added a `raise ValueError` on [L289](https://github.com/santisoler/harmonica/blob/6e6e47b76ff448a97e9afc6ef7fa04d6d3bc6bfb/harmonica/forward/tesseroid.py#L289) and Numba does not complain about it.
I'm proposing something like that inside `_adaptive_discretization`.

santisoler:
@leouieda I've made the last changes we discussed:
- added a gallery example,
- moved all tests inside `harmonica/tests` folder,
- created an alias for `pytest.mark.skipif()` decorator on `tests/utils.py`,
- edited the `Makefile` to generate two targets for testing: with Numba enabled and disabled,
- refactor the spherical shell comparison: instead of computing `diff < 0.1` with `npt.assert_allclose`,
- add test function for testing `_longitude_continuity` on lower level,
- and fixed linting errors.

I must admit I've disabled some pylint errors, specially `too-many-locals`, `too-many-arguments` and `invalid-name` for indexes, mainly on jitted functions. Please check if you agree with those disabled pylint warnings.

If you cannot find any big issue with this code, I think we could merge it and open issues for future enhancements or refactors.

santisoler:
@leouieda AppVeyor complains when running `python -c "import harmonica; harmonica.test()"` because the env variable "NUMBA_DISABLE_JIT" is not defined.

On the last commit I made the `@requires_numba` decorator to raise an `EnvironmentError` if `NUMBA_DISABLE_JIT` is not explicitly set. I prefer to check if it's defined instead of hardcoding a specific decision in case it isn't. What do you think?

This raised two questions:
1. Haven't we moved from AppVeyor to Azure? https://github.com/fatiando/contributing/pull/7
2. Now we run two tests (with and without Numba) should we modify the `harmonica.test()` function? Keep in mind that we will need to set `NUMBA_DISABLE_JIT` beforehand.

harmonica/tests/utils.py
```diff
@@ -0,0 +1,37 @@
...
+    want to run if Numba jit is disabled.
+    The decorated test functions will be run and checked if pass or fail, but won't be
+    taken into account for meassuring coverage.
+    If the test function will run Numba code, but doesn't involve great computational
+    load, we reccomend using the ``@pytest.mark.use_numba`` instead. Therefore the test
+    function will be run twice: one with Numba jit enabled, and another one with Numba
+    jit disable to check coverage.
+    """
+    # Check if env variable NUMBA_DISABLE_JIT is defined
+    if os.getenv("NUMBA_DISABLE_JIT") is None:
```
leouieda:
```suggestion
    if os.environ.get("NUMBA_DISABLE_JIT", None) is None:
```

santisoler:
> @leouieda On the last commit I made the `@require_numba` decorator to raise an `EnvironmentError` if `NUMBA_DISABLE_JIT` is not explicitly set. I prefer to check if it's defined instead of hardcoding a specific decision in case it isn't. What do you think?
> 
> On the other hand, AppVeyor complains when running `python -c "import harmonica; harmonica.test()"` because the env variable "NUMBA_DISABLE_JIT" is not defined.
> Maybe we could solve #63 and then merge this PR.
> 
> Nevertheless, it raised one question:
> 
>     * Now we run two tests (with and without Numba) should we modify the `harmonica.test()` function? Keep in mind that we will need to set `NUMBA_DISABLE_JIT` beforehand.

We decided to assume that Numba jit is enabled if the env variable `NUMBA_DISABLE_JIT` is not defined.

Besides, it's ok that `python -c "import harmonica; harmonica.test()"` runs with Numba jit enabled, because it doesn't check coverage, it must test if everything is working as expected on the user installation.

PR #61, opened 2019-05-08
Refactor coordinates conversion

santisoler:
Geographic coordinates conversion functions now get the full coordinates
as arguments rather than only the `latitude` and `radius`/`height`, and returns
the full set of converted coordinates.
The previous implementation was inconvenient for real applications.

**Reminders**

### DIFF
--- a/harmonica/coordinates.py
+++ b/harmonica/coordinates.py
@@ -6,7 +6,7 @@ import numpy as np
 from . import get_ellipsoid
 
 
-def geodetic_to_spherical(coordinates):
+def geodetic_to_spherical(longitude, latitude, height):
     """
     Convert from geodetic to geocentric spherical coordinates.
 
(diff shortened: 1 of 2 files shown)


leouieda:
@santisoler just noticed that if we want these functions to work with Verde gridders, they should actually take `longitude, latitude, height` instead of `coordinates`: https://github.com/fatiando/verde/blob/master/verde/base/base_classes.py#L294

For example, if we make a spherical equivalent layer using Verde, we could use these functions to generate geodetic grids easily by passing them in as the `projection` argument to the `grid` method.

I don't think it's too inconvenient since the user code would change from `geodetic_to_spherical(coordinates)` to `geodetic_to_spherical(*coordinates)`.

santisoler:
I'm a little bit confused about your proposal.

> just noticed that if we want these functions to work with Verde gridders

On what cases you want Verde gridders to convert from spherical to geodetic or from geodetic to spherical?
Maybe I'm missing an obvious example for that...

The `coordinates` tuple on the line you linked is expected to have only `longitude`, `latitude` coordinates, not `longitude`, `latitude`, `height`. Remeber that the `height`/`radius` array is crucial for the spherical to geodetic conversion. I feel I'm missing something around this...

leouieda:
The use case I had in mind what something like this:

```python
# Load observed gravity in geodetic coordinates
lon, lat, height, gravity = np.loadtxt(...)

# Convert to geocentric spherical so we can use a spherical gridder
coordinates = hm.geodetic_to_spherical((lon, lat, height))

# Grid using equivalent sources in spherical coordinates
eql = hm.EQLHarmonic(kernel="spherical", ...)
eql.fit(coordinates, gravity)

# Generate a geodetic grid using our spherical gridder
grid = eql.grid(region=(...), spacing=0.1, extra_coords=[1000],
                projection=hm.spherical_to_geodetic)
# extra_coords is the ellipsoidal height of the grid
```

In this case, we can use a spherical gridder to generate a geodetic grid without much effort. This is the same way we can make geodetic grids from Cartesian gridders in Verde.

The expectation of `coordinates=(latitude, longitude)` is only in the docstring. The code itself doesn't care as long as the number of coordinates `projection` accepts is the same as the number `grid_coordinates` returns. Using `extra_coords` will make it return 3 instead of 2.

santisoler:
Thanks for that example @leouieda.
Now I understand what you meant.
I agree and just made those changes.
I think it's ready to be merged.

PR #74, opened 2019-07-19
Simplify title of tesseroid gallery example

### LINKED ISSUE #73: Remove "Forward modelling" from tesseroid gallery example

Because all gallery examples that perform forward modelling now live inside `forward`, it's redundant to add "Forward modelling" to the example title. Would be nice to remove it and title it as "Tesseroid" or "Single Tesseroid".

santisoler:
Change the title of the tesseroid gallery example by removing the "Forward Modelling"
part because it's explicit on the title of the examples section where it lives.

Fixes #73 

**Reminders**


examples/forward/tesseroid.py
```diff
@@ -1,6 +1,6 @@
 """
-Tesseroid Forward Modelling
-===========================
+Single Tesseroid
```
leouieda:
```suggestion
Tesseroids (spherical prisms)
```

santisoler: Yes, I like it. I was thinking that in the future other tesseroids examples may appear on the gallery, such as Tesseroid Mesh. But I think the difference between titles would be on that "Meshes", so the "Single" part on this one is irrelevant. And adding "spherical prisms" is nice, I tend to forget no everyone is familiar with the tesseroid term.

PR #71, opened 2019-07-17
Forward modelling for point masses in Cartesian coordiantes

santisoler:
Extend point_mass_gravity() to compute gravitational fields of point masses in Cartesian
coordinates. Add needed Numba jitted kernel functions in Cartesian coordinates. Add
coordinate_system argument to point_mass_gravity(). Add tests for the extended feature.

Fixes #48 

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to the docstring, gallery, and/or tutorials.

PR #78, opened 2019-07-22
Add equivalent layer for harmonic functions

### LINKED ISSUE #39: General equivalent layer solution

The equivalent layer (EQL) methods are great for interpolation, data integration, upward continuation, etc. Our implementations will be based on the Verde gridding classes. The most basic kind would be an EQL that uses 1/r as the basis function. This can be used with any kind of harmonic function but cannot integrate different fields (grav and gradients) and can't reduce to the pole. But it fits the Verde model well. We should implement this as a first step and then make more specialized gridders that can receive multiple fields as input. 

The class should inherit from `verde.base.BaseGridder` and follow its protocols. It could live in a new package called `harmonica/equivalent_layer/harmonic.py` or something like that.

santisoler:
Add an Equivalent Layer gridder for harmonic functions, based on
verde.BaseGridder using 1/r as the basis function. By default it puts
one point source bellow each observation point at a constant relative
depth and fits the corresponding coefficient to each source. Also
a custom set of point sources can be passed. Add a gallery example
showing how to grid a small region of the magnetic anomaly data from Rio
de Janeiro.

Fixes #39 

**Reminders**


santisoler:
I want to use some functions from #71 when predicting interpolated values and computing distances in Cartesian coordinates. They must be incorporated after #71 is merged.
I'm still using the term "vertical" for specifying the downward coordinates. It should be changed in agreement with the term used in #71.

santisoler:
It also uses some functions from the `dev` version of Verde. A new Verde release should be made in order to CIs to pass.

harmonica/equivalent_layer/harmonic.py
```diff
@@ -0,0 +1,228 @@
...
+        a default set of points will be created putting a single point source bellow
+        each observation point at a depth proportional to the mean distance to the
+        nearest k observation points [Cooper2000]_. Default None.
+    depth_factor : float (optional)
+        Adimensional factor to set the depth of each point source.
+        If ``points`` is None, a default set of point will be created putting
+        a single point source bellow each obervation point at a depth given by the
+        product of the ``depth_factor`` and the mean distance to the nearest
+        k obervation points. A greater ``depth_factor`` will increase the depth of the
+        point source. This parameter is ignored if ``points`` is not None. Default 3.
```
leouieda: Add citation to Cooper for the depth factor.

examples/eql_harmonic.py
```diff
@@ -0,0 +1,64 @@
...
+import verde as vd
+import numpy as np
+import pyproj
+import cartopy.crs as ccrs
+import matplotlib.pyplot as plt
+from sklearn.model_selection import ShuffleSplit
+
+
+# Fetch magnetic anomaly data from Rio de Janeiro
+data = hm.datasets.fetch_rio_magnetic()
```
santisoler: Get a smaller region on the magnetic anomaly data instead of downsampling it. Try to catch a dipole.

examples/eql_harmonic.py
```diff
@@ -0,0 +1,64 @@
...
+# Project coordinates
+projection = pyproj.Proj(proj="merc", lat_ts=data.latitude.mean())
+data["easting"], data["northing"] = projection(
+    data.longitude.values, data.latitude.values
+)
+coordinates = (data["easting"], data["northing"], data.altitude_m)
+
+# Perform a cross-validation in order to score the interpolator
+gridder = hm.EQLHarmonic()
+gridder.set_params(depth_factor=1)
```
santisoler: Try to increase the `k_nearest` because we are dealing with airborne magnetic data and nearest observation points may be too close.

leouieda: Better to pass `depth_factor` to the constructor.

examples/eql_harmonic.py
```diff
@@ -0,0 +1,64 @@
...
+shuffle = ShuffleSplit(n_splits=10, test_size=0.3, random_state=1)
+scores = vd.cross_val_score(
+    gridder, coordinates, data.total_field_anomaly_nt, cv=shuffle
+)
```
leouieda: Put a comment here explaining what the score is and what it means.

harmonica/equivalent_layer/harmonic.py
```diff
@@ -0,0 +1,228 @@
...
+        self
+            Returns this estimator instance for chaining operations.
+        """
+        coordinates, data, weights = check_fit_input(coordinates, data, weights)
+        # Capture the data region to use as a default when gridding.
+        self.region_ = get_region(coordinates[:2])
+        coordinates = n_1d_arrays(coordinates, 3)
+        if self.points is None:
+            # Put a single point source bellow each observation point at a depth three
+            # times the meadian distance to the nearest k observation points.
```
santisoler:
```suggestion
            # times the median distance to the nearest k observation points.
```

harmonica/equivalent_layer/harmonic.py
```diff
@@ -0,0 +1,228 @@
...
+            point_east, point_north, point_upward = tuple(i.copy() for i in coordinates)
+            point_upward += self.depth_factor * median_distance(
```
leouieda: After #92 is merged this will need to be a `-`

santisoler: After we merge #92, the plus sign must be a minus in order to actually put the point sources below the observation points.
```suggestion
            point_upward -= self.depth_factor * median_distance(
```

santisoler:
@leouieda I updated this branch according to what we discussed yesterday. I like how the gallery example looks now, in fact it achieved a score of 0.98 :tada:.
Let me know if you feel we should change anything else.

leouieda:
Hi Santi, I ran you example now but it seems a bit over-sampled to me. Looking closer, there are a lot of spikes along the flight lines that we probably due to the large number of sources. I tried messing with the region a bit but didn't have time to find a good fit. You might need to make the sources deeper (that static shift we talked about) or do a blockmean on the data points to downsample (it's better than taking every other point). The grid spacing is also a bit too small for the final grid since the lines are separated by around 2500m.

leouieda:
EQLs are hard. Don't be dismayed! It takes some practice and it's a really hard problem. I wouldn't trust that high score too much as well. You might want to make another example with synthetic data (or maybe start a tutorial page using synthetic to be more controlled and leave the gallery with real data).

examples/eql_harmonic.py
```diff
@@ -0,0 +1,74 @@
...
+interpolator. We will use :class:`harmonica.EQLHarmonic` to generate a set of point
+sources beneath the observation points that predicts the observed data. Then these point
+sources will be used to interpolate the data values into a regular grid at a constant
+height.
+"""
+import harmonica as hm
+import verde as vd
+import numpy as np
+import pyproj
+import cartopy.crs as ccrs
```
santisoler:
```suggestion
```

leouieda:
I'm experimenting with this a bit and you probably won't have to change the EQL implementation. I'll see if I can come up with another example.

doc/api/index.rst
```diff
@@ -16,6 +16,14 @@ Gravity Corrections
     normal_gravity
     bouguer_correction
 
+Gridding and Interpolation
```
leouieda:
```suggestion
Equivalent Layers
```

The EQLs can do much more than interpolation: upward-continuation, data integration, reduction to the pole, derivatives, etc.

leouieda:
@santisoler I've been experimenting with the example and code a bit.

I isolated another anomaly that was better for visualizing. This is the result using the same parameters you were using but with a diverging colormap with 0 centered on white:

meh

Notice that there are a lot of artifacts from the flight lines. The solution is also very sensitive to the `depth_factor` and `k_nearest` arguments.

These are the results with a larger grid spacing (400m) and a uniform depth (500m below data):

meh

Notice that the stripes are gone. This due mainly to the larger spacing. The single depth is also easier to understand. I think the variable depth might be useful in other circumstances but probably not in the cases we're looking at right now. So it might be better to use the simpler case first and add the option of variable depth as we progress.

What do you think? 

If you agree, I'll go ahead and commit my changes to the code and example.

santisoler:
Hi @leouieda! I really like those results and agree to keep things simple and try more complex layouts in the future. Feel free to commit your changes!

leouieda:
@santisoler I pushed my modifications but it still needs some polish. I haven't checked the tests, the docstrings need some updating and polish, and the example plot could be nicer.

leouieda:
I'll take a look at the prism PR tomorrow.

leouieda:
@santisoler this is ready for a review. Here is what I've done:

* Placed EQL examples in their own category. This one is about gridding and upward continuation.
* Add Dampney 1969 citation
* Remove the variable depth code to simplify this setup
* Rework the testing to check interpolating from one grid to a denser grid
* Example uses another anomaly, no cross-validation (to make things simpler and because we don't know if it actually works), balances the color maps, uses a larger spacing, and more text at the top explaining what EQLs do.
* Class docstring lists what this EQL is good for and what it can't do. This will be useful since other layers can so other things and we can link them together this way.

leouieda:
@birocoles it would be great if you took a look at this as well. Particularly the `harmonica/equivalent_layer/harmonic.py` and `examples/eql/harmonic.py`. I don't think we said anything terribly wrong but you're the real expert here :man_student:.

.azure-pipelines.yml
```diff
@@ -92,6 +92,12 @@ jobs:
   - bash: echo "##vso[task.prependpath]$CONDA/bin"
     displayName: Add conda to PATH
 
+  # On Hosted macOS, the agent user doesn't have ownership of Miniconda's installation
```
santisoler: Shouldn't we add these lines on a different PR?

leouieda:
> I give you one point for showing both the capabilities and the flaws of the gridder.

:1st_place_medal: :confetti_ball: 

> I would make clearer what the depth argument means: at first sight it seems that all point sources are located at a constant depth by default, but in fact they are set at a constant relative depth beneath each observation point. What do you think?

I completely agree. Would you mind adding this? Feel free to merge once that's done. We can always improve the docs but gotta get it out at some point ;-)

> Shouldn't we add these lines on a different PR?

You caught me being lazy... :disappointed: I'll open the PR.

santisoler:
I renamed the `depth` argument to `relative_depth` and improved the docstrings to make it clearer what this argument represents. I've also add a little comment on the gallery example explaining this.
I'll merge this if CIs agree. I agree we can change things later.

PR #81, opened 2019-07-24
Unify the computation of distances in Cartesian and spherical coordinates

### LINKED ISSUE #69: Unify computation of distances in spherical and Cartesian coordinates

To prevent repeated code, would be nice to have single functions for computing distances between two points in both spherical and Cartesian coordinates. They must be Numba jitted functions to speed things up.

For the spherical coordinates case, the points are given in `longitude`, `latitude` and `radius` coordinates, but several trigonometric functions must be applied in order to ultimately get the distance between these two points. Some forward models precompute these quantities in order to save time. So, would be nice to have a private function that takes all these precomputed quantities as arguments and returns the distance, and a public function that just takes the points' coordinates, compute this quantities and use this private function to get the distance between them.

Moreover, some forward models just needs the distance square, so we could make another private function that just returns the distance square, i.e. doesn't apply the square root.

For example:
```python
@jit(nopython=True)
def distance_spherical(point_a, point_b):
    # Get coordinates of the two points
    longitude, latitude, radius = point_a[:]
    longitude_p, latitude_p, radius_p = point_b[:]
    # Convert angles to radians
    longitude, latitude = np.radians(longitude), np.radians(latitude)
    longitude_p, latitude_p = np.radians(longitude_p), np.radians(latitude_p)
    # Compute trigonometric quantities
    cosphi_p = np.cos(latitude_p)
    sinphi_p = np.sin(latitude_p)
    cosphi = np.cos(latitude)
    sinphi = np.sin(latitude)
    return _distance_spherical(
        longitude, cosphi, sinphi, radius, longitude_p, cosphi_p, sinphi_p, radius_p
    )


@jit(nopython=True)
def _distance_spherical(
    longitude, cosphi, sinphi, radius, longitude_p, cosphi_p, sinphi_p, radius_p
):
    return np.sqrt(
        _distance_sq_spherical(
            longitude, cosphi, sinphi, radius, longitude_p, cosphi_p, sinphi_p, radius_p
        )
    )


@jit(nopython=True)
def _distance_sq_spherical(
    longitude, cosphi, sinphi, radius, longitude_p, cosphi_p, sinphi_p, radius_p
):
    coslambda = np.cos(longitude_p - longitude)
    cospsi = sinphi_p * sinphi + cosphi_p * cosphi * coslambda
    return (radius - radius_p) ** 2 + 2 * radius * radius_p * (1 - cospsi)
```

santisoler:
Add functions to compute distances in both Cartesian and spherical coordinates on
forward/utils.py. Make point mass and tesseroid forward models to use these functions.

Fixes #69 

**Reminders**

### DIFF
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -24,7 +24,6 @@ Forward modelling
 
     point_mass_gravity
     tesseroid_gravity
-    distance
 
 Isostasy
 --------
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -24,6 +24,8 @@ Forward modelling
 
     point_mass_gravity
     tesseroid_gravity
+    distance_cartesian
+    distance_spherical
 
 Isostasy
 --------
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -24,8 +24,7 @@ Forward modelling
 
     point_mass_gravity
     tesseroid_gravity
-    distance_cartesian
-    distance_spherical
+    distance
 
 Isostasy
 --------
--- a/harmonica/forward/point_mass.py
+++ b/harmonica/forward/point_mass.py
@@ -6,7 +6,7 @@ from numba import jit
 
 from ..constants import GRAVITATIONAL_CONST
 from .utils import (
-    _check_coordinate_system,
+    check_coordinate_system,
     distance_cartesian,
     distance_spherical,
     distance_spherical_core,
@@ -147,7 +147,7 @@ def point_mass_gravity(
         "spherical": {"potential": kernel_potential_spherical, "g_r": kernel_g_r},
     }
     # Sanity checks for coordinate_system and field
-    _check_coordinate_system(coordinate_system)
+    check_coordinate_system(coordinate_system)
     if field not in kernels[coordinate_system]:
         raise ValueError("Gravity field {} not recognized".format(field))
     # Figure out the shape and size of the output array
--- a/harmonica/forward/point_mass.py
+++ b/harmonica/forward/point_mass.py
@@ -5,6 +5,11 @@ import numpy as np
 from numba import jit
 
 from ..constants import GRAVITATIONAL_CONST
+from .utils import (
+    DISTANCE_CARTESIAN_NUMBA,
+    _distance_sq_cartesian,
+    _distance_sq_spherical,
+)
 
 
 def point_mass_gravity(
@@ -199,7 +204,7 @@ def kernel_potential_cartesian(easting, northing, down, easting_p, northing_p, d
     """
     Kernel function for potential gravity field in Cartesian coordinates
     """
-    return 1 / _distance_cartesian(
+    return 1 / DISTANCE_CARTESIAN_NUMBA(
         [easting, northing, down], [easting_p, northing_p, down_p]
     )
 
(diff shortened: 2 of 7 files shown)


santisoler:
@leouieda I'm thinking we could merge `distance_spherical()` and `distance_cartesian()` into a single `distance()` function that takes `coordinate_system` as an argument, as `point_mass_gravity()` does. What do you think?

leouieda:
:+1: yes! API consistency :fireworks:

santisoler:
@leouieda I merged the computation of distances into a single public function. I've also added a `_check_coordinate_system()` to unify the verification of valid coordinate system. It's currently used on `point_mass_gravity()` and `distance()`.

I had to make little changes on the private jitted functions because they are needed on the forward modelling functions.

Let me know if you see something that could be improved.

doc/api/index.rst
```diff
@@ -24,6 +24,7 @@ Forward modelling
 
     point_mass_gravity
     tesseroid_gravity
+    distance
```
leouieda: I think this probably shouldn't be in the public AP if it's a utility meant for use in the forward modeling modules. We can always make it public later but removing the function is harder.

harmonica/forward/point_mass.py
```diff
@@ -303,8 +287,8 @@ def kernel_g_r(
...
     Kernel function for radial component of gravity gradient in spherical coordinates
     """
-    coslambda = np.cos(longitude_p - longitude)
-    cospsi = sinphi_p * sinphi + cosphi_p * cosphi * coslambda
-    distance_sq = (radius - radius_p) ** 2 + 2 * radius * radius_p * (1 - cospsi)
+    distance_sq, cospsi, _ = _distance_sq_spherical_core(
+        longitude, cosphi, sinphi, radius, longitude_p, cosphi_p, sinphi_p, radius_p
+    )
     delta_z = radius_p * cospsi - radius
     return delta_z / distance_sq ** (3 / 2)
```
leouieda: I learned a while ago that sometimes this is actually a lot slower than `np.sqrt(distance)**3`. So we don't really need the distance to be squared.

santisoler:
> Hi Santi, I've been going through the code and it seems like the `distance` function isn't really used anywhere. The actual functions used are the jit functions. So these should not have `_`. The `_` indicates functions that aren't meant to be used outside of their module. 

Hi  @leouieda. Thanks for the review. I tend to get confused about *private* and *public* functions and if all public functions should be in the API. Thanks for making it clearer.

> And these are used across `forward`. Here is my proposal:
> 
>     1. Remove the `distance` function from the API. It isn't used right now and we can always add it later if necessary. No need to delete it but it's best to wait for a use case before making it public.

No problem!

>     2. Make the jit function "public" but not in the API.

I agree with this.

>     3. Put the `srqt` in the distances. Squaring something is a cheap operation and pretty much all functions do the sqrt at some point.

You are right, working the squared values will eventually create some mistakes.

harmonica/forward/point_mass.py
santisoler: :+1: Didn't know that. I thought that applying a single power to a value would be faster than applying two different ones in series. Although I have no references to hold this statement, just intuition.

leouieda:
:+1: good to merge on my part. Looks great!

PR #92, opened 2019-08-07
Change vertical direction in Cartesian coordinate pointing upwards

### LINKED ISSUE #89: Standarize Cartesian coordinates with z pointing up

After a discussion, @leouieda and I agreed that would be better to standarize the vertical coordinate in Cartesian coordinate system pointing upwards.
The benefits of this decision are:
- Computational heights would be positive values, saving users from converting their height coordinates of observation points to negative values.
- The same goes for structures above the zero height, like topography. Having negative topography heights is not intuitive.
- When specifying some source geometries (like prisms), their boundaries can be standardized as the ones in spherical coordinates are defined: `w`, `e`, `s`, `n`, `bottom`, `top`. If the vertical direction points upwards, the `bottom` value will be numerically lower than the `top` one, in the same way as tesseroids are defined. 

One of the drawbacks of this idea is that the coordinate system would be left-handed. Nevertheless, this can be solved with good documentation and careful implementation of the math beneath the forward models.

This new idea also change how gravity acceleration is defined. For sake of standarization, we propose that the forward models that computes the vertical (or radial) component of the gravity acceleration return the downward component rather than the one oriented according to the new vertical direction. We agreed on this because among geophysicists the gravity is always expected to be pointing downward, thus a positive density contrast produces a positive anomaly.
This should be well specified on the documentation and on the future page about coordinates systems (#68).

santisoler:
Change the direction of the Cartesian vertical coordinate: it points upwards instead of downwards.
The calculation of the g_z field returns the downward component of the gravity acceleration both for point masses in spherical and Cartesian coordinates and also for tesseroids.

Fixes #89 

**Reminders**


examples/forward/point_mass.py
```diff
@@ -21,14 +21,14 @@
 # Define two point masses with oposite mass values of 10000000 kg
 easting = [5e3, 15e3]
 northing = [5e3, 15e3]
-down = [7e3, 2.5e3]
-points = [easting, northing, down]
+upward = [-7e3, -2.5e3]
```
leouieda:
```suggestion
# The vertical coordinate is positive upward so negative numbers represent depth
upward = [-7e3, -2.5e3]
```

Issue #70, opened 2019-07-17
Numba deprecation of reflected list


santisoler:
**Description of the problem**

Pytest is currently raising a `NumbaPendingDeprecationWarning` due to a scheduled deprecation of type `reflected list`.
I believe it's raised when a list is passed as argument to a jitted function.

**Full code that generated the error**

```bash
make test
```

**Full error message**

```
# Run a tmp folder to make sure the tests are run on the installed version
mkdir -p tmp-test-dir-with-unique-name
cd tmp-test-dir-with-unique-name; NUMBA_DISABLE_JIT=1 MPLBACKEND='agg' pytest --cov-config=../.coveragerc --cov-report=term-missing --cov=harmonica --doctest-modules -v --pyargs harmonica
====================================== test session starts ======================================
platform linux -- Python 3.7.3, pytest-5.0.1, py-1.8.0, pluggy-0.12.0 -- /home/santi/.miniconda3/envs/harmonica/bin/python
cachedir: .pytest_cache
rootdir: /home/santi/documentos/geo/git/harmonica
plugins: cov-2.7.1
collected 72 items                                                                              

../harmonica/coordinates.py::harmonica.coordinates.geodetic_to_spherical PASSED           [  1%]
../harmonica/coordinates.py::harmonica.coordinates.spherical_to_geodetic PASSED           [  2%]
../harmonica/ellipsoid.py::harmonica.ellipsoid.ReferenceEllipsoid PASSED                  [  4%]
../harmonica/ellipsoid.py::harmonica.ellipsoid.get_ellipsoid PASSED                       [  5%]
../harmonica/ellipsoid.py::harmonica.ellipsoid.print_ellipsoids PASSED                    [  6%]
../harmonica/ellipsoid.py::harmonica.ellipsoid.set_ellipsoid PASSED                       [  8%]
../harmonica/forward/tesseroid.py::harmonica.forward.tesseroid.tesseroid_gravity PASSED   [  9%]
../harmonica/tests/test_coordinates.py::test_geodetic_to_spherical_with_spherical_ellipsoid PASSED [ 11%]
../harmonica/tests/test_coordinates.py::test_geodetic_to_spherical_on_equator PASSED      [ 12%]
../harmonica/tests/test_coordinates.py::test_geodetic_to_spherical_on_poles PASSED        [ 13%]
../harmonica/tests/test_coordinates.py::test_spherical_to_geodetic_with_spherical_ellipsoid PASSED [ 15%]
[...]

PR #96, opened 2019-09-07
WIP Add horizontal Cartesian components for point masses

### LINKED ISSUE #85: Add computation of horizontal components of gravitational acceleration

Point mass and tesseroid forward models are currently capable of computing only the potential and the vertical (or radial) component of the gravitational acceleration.
Would be nice to add the ability to compute the horizontal componentes, i.e. the easting and northing ones.

birocoles:
Implement horizontal Cartesian components `g_northing`and `g_easting` for point masses. This PR addresses some of #85 . Components in spherical coordinate system will be implemented in the future. 

**Reminders**


birocoles:
Hi @santisoler , 

Please take a look at this PR and give me a feedback. I didn't include the routines in the `doc/api/index.rst` nor create a tutorial **yet**.

harmonica/forward/point_mass.py
```diff
@@ -189,6 +193,10 @@ def point_mass_gravity(
     # Convert to more convenient units
     if field == "g_z":
         result *= 1e5  # SI to mGal
+    if field == "g_northing":
```
santisoler: There's no need for creating one `if` statement for each component. The units conversion can be summarized as follows:
```python
if field in ("g_easting", "g_northing", "g_z"):
    result *= 1e5  # SI to mGal
```

harmonica/forward/point_mass.py
```diff
@@ -249,6 +257,32 @@ def kernel_g_z_cartesian(easting, northing, upward, easting_p, northing_p, upwar
...
+def kernel_g_northing_cartesian(
+    easting, northing, upward, easting_p, northing_p, upward_p
+):
+    """
+    Kernel function for northing component of gravity gradient in Cartesian coordinates
+    """
+    distance = distance_cartesian(
+        [easting, northing, upward], [easting_p, northing_p, upward_p]
+    )
+    return (northing - northing_p) / distance ** 3
```
santisoler: There's a little sign mistake on this line, but I think I triggered it when I wrote the `kernel_g_z_cartesian`.

Because the acceleration vector points towards the point source (if it has positive mass), this line should be:
```suggestion
    return - (northing - northing_p) / distance ** 3
```

Consider a point mass located on the origin and two computation points located at 10 meters on the northing direction, but located at opposite sides of the point source. The `g_northing` on the computation point located on the positive axe of `northing` should be negative, while `g_northing` on the other one should positive (and both with the same absolute value).
So, the following code should give the following result:
```python
import harmonica as hm
coordinates = ([0, 0], [-10, 10], [0, 0])
point_mass = [0, 0, 0]
masses = [100, 100]
hm.point_mass_gravity(
    coordinates, point_mass, masses, field="g_northing", coordinate_system="cartesian"
)
>>> array([6.673e-06,  -6.673e-06])
```

I think this mistake is my fault because I haven't added a comment on `kernel_g_z_cartesian` explaining why I omit the minus sign. The reason for this omission is that `g_z` represent the downward component, although our vertical axes points upwards. Therefore, we must remove the minus sign in order to get the opposite component of the acceleration vector.

I think it would be nice to check the sign of the acceleration components on the symmetry tests (not only for the new `g_easting` and `g_northing`, but also for the `g_z` symmetry) and would be nice to add a little comment on the `kernel_g_z_cartesian` function making very clear that we are intentionally omitting the minus sign. What do you think? Would you like to add these two things on this PR?

harmonica/forward/point_mass.py
```diff
@@ -249,6 +257,32 @@ def kernel_g_z_cartesian(easting, northing, upward, easting_p, northing_p, upwar
...
+def kernel_g_easting_cartesian(
+    easting, northing, upward, easting_p, northing_p, upward_p
+):
+    """
+    Kernel function for easting component of gravity gradient in Cartesian coordinates
+    """
+    distance = distance_cartesian(
+        [easting, northing, upward], [easting_p, northing_p, upward_p]
+    )
+    return (easting - easting_p) / distance ** 3
```
santisoler: Same for the `kernel_g_northing_cartesian`:
```suggestion
    return - (easting - easting_p) / distance ** 3
```

harmonica/forward/point_mass.py
birocoles: Perfect! I would suggest this modification in the future. Thank you

harmonica/forward/point_mass.py
birocoles: Ok @santisoler , I will fix these problems. They were my mistake. I will also include a comment in the gz about this signal convention.

harmonica/forward/point_mass.py
birocoles: ok

harmonica/forward/point_mass.py
birocoles: In the tests for verifying the signals of the components `g_z`, `g_northing` and `g_easting`, may I use simple `assert field < 0`, `assert > 0` and `assert == 0` commands? I did not find routines to do this kind of assertion in the `numpy.testing`.

santisoler: The `assert field < 0` and `assert field > 0` are valid! We try to avoid the `assert field == 0` because on some numerical computations the result is very close to zero but not necessarily strictly equal to zero. So, you can use `assert field == 0` only if we are sure that the computed result will be strictly zero. For example, results from analytical solutions tend to satisfy this requirement, but on numerical approximations having a result strictly equal to zero is very rare.

Nevertheless, on this case I would like to take into account the sign of the mass instead of hardcoding the expected sign of the field. So, for example, if you put a single point mass with positive mass, then the g_z value above it should be positive, so we can assert if the sign of the g_z matches the sign of the mass:

```python
assert np.sign(mass) == np.sign(g_z)
```

If the computation point is now bellow the point mass, then the `g_z` should have the opposite sign of the mass:


```python
assert np.sign(mass) == - np.sign(g_z)
```

Do you agree?

birocoles: Perfect!

leouieda: @birocoles to check if a float is close to 0, use `numpy.testing.assert_allclose(number, 0)`

birocoles: I think tests abotu signal are ok now!

birocoles: I have also included a test for comparing numerical derivatives of Cartesian `potential` and the analytical components given by `g_z`, `g_northing` and `g_easting`. Everything is ok for the horizontal components. There is a problem in `g_z`. During this test, I verified that `kernel_potential_cartesian` was using `distance_spherical` instead of `distance_cartesian`. I fixed this problem, but the problem with the test is still there.

birocoles:
I have just fixed the problem with numerical derivatives. There is still a problem related to the unused `distance_spherical` imported from utils at the begining of `point_mass.py`.

birocoles:
Hi @santisoler and @leouieda , I have removed the unused `distance_spherical` function from `point_mass_gravity.py`. The functions computing things in spherical coordinates use the function `distance_spherical_core`. Besides, I have included the components `g_northing`, `g_easting` and `g_upward` in the documentation. I opted for using `g_upward` instead of `g_up` for convenience. No problem if you prefer a short name instead. I have also replaced the terms `gravity` by `gravitational`.

harmonica/forward/point_mass.py
```diff
@@ -42,22 +37,32 @@ def point_mass_gravity(
...
 
     .. math::
 
-        \vec{g} = \nabla V.
+        \vec{g} = \nabla V
 
-    Therefore, the :math:`z` component of :math:`\vec{g}` at the point :math:`P` can be
-    computed as (remember that :math:`z` points upwards):
+    and has components :math:`g_{northing}(P)`, :math:`g_{easting}(P)` and
+    :math:`g_{upward}(P)` given by
```
santisoler: We could try using LaTeX's `\text` in order to make the subscripts look like text rather than variables.

harmonica/forward/point_mass.py
```diff
@@ -231,30 +240,61 @@ def kernel_potential_cartesian(
     easting, northing, upward, easting_p, northing_p, upward_p
 ):
     """
-    Kernel function for potential gravity field in Cartesian coordinates
+    Kernel function for potential gravitational field in Cartesian coordinates
     """
-    return 1 / distance_spherical(
-        (easting, northing, upward), (easting_p, northing_p, upward_p)
+    distance = distance_cartesian(
```
santisoler: Thanks for catching this! This was an awful mistake! :man_facepalming: 
When I see this kind of mistakes I remind myself the importance of reviewing each other's code.

harmonica/forward/point_mass.py
```diff
@@ -231,30 +240,61 @@ def kernel_potential_cartesian(
...
+        [easting, northing, upward], [easting_p, northing_p, upward_p]
```
santisoler: Is there a special reason why you replaced tuples with lists?
We try to avoid passing lists to Numba jitted functions because Numba has deprecated list reflection (see #70 and [this Numba doc entry](https://numba.pydata.org/numba-doc/latest/reference/deprecation.html?highlight=reflected#deprecation-of-reflection-for-list-and-set-types) for more information).
Is there any deprecation warning raise when running the tests?

harmonica/forward/point_mass.py
```diff
@@ -231,30 +240,61 @@ def kernel_potential_cartesian(
...
     """
-    Kernel function for downward component of gravity gradient in Cartesian coordinates
+    Kernel function for downward component of gravitational gradient in Cartesian coordinates
     """
     distance = distance_cartesian(
         [easting, northing, upward], [easting_p, northing_p, upward_p]
     )
+    # Remember that the ``g_z`` field returns the downward component of the
+    # gravitational acceleration. As a consequence, it is multiplied by -1. Notice that
+    # the ``g_z`` does not have the minus signal observed at the compoents
```
santisoler:
```suggestion
    # the ``g_z`` does not have the minus signal present on the
```

harmonica/forward/point_mass.py
```diff
@@ -231,30 +240,61 @@ def kernel_potential_cartesian(
...
+    # ``g_northing`` and ``g_easting``.
```
santisoler:
```suggestion
    # ``g_northing`` and ``g_easting`` components.
```

harmonica/tests/test_point_mass.py
```diff
@@ -91,6 +170,119 @@ def test_g_z_symmetry():
     npt.assert_allclose(results[0], -results[1])
 
 
+@pytest.mark.use_numba
+def test_g_z_signal():
```
santisoler: Shouldn't this be called `test_g_z_sign` instead of `test_g_z_signal`?
```suggestion
def test_g_z_sign():
```

harmonica/tests/test_point_mass.py
```diff
@@ -91,6 +170,119 @@ def test_g_z_symmetry():
...
+    """
+    Test if g_z field of a positive point mass has the correct signal
```
santisoler: Same here:
```suggestion
    Test if g_z field of a positive point mass has the correct sign
```

harmonica/tests/test_point_mass.py
```diff
@@ -91,6 +170,119 @@ def test_g_z_symmetry():
...
+    """
+    # Define a single point mass
+    point_mass = [-10, 100.2, -300.7]
+    mass = [2670]
+    # Define three computation points located above, at the same depth and
+    # bellow the point mass
+    easting = np.zeros(3)
+    northing = np.zeros(3) + 52.3
+    z = np.array([100.11, -300.7, -400])
```
santisoler:
```suggestion
    upward = np.array([100.11, -300.7, -400])
```

harmonica/tests/test_point_mass.py
```diff
@@ -91,6 +170,119 @@ def test_g_z_symmetry():
...
+    coordinates = [easting, northing, z]
```
santisoler:
```suggestion
    coordinates = [easting, northing, upward]
```

harmonica/tests/test_point_mass.py
```diff
@@ -91,6 +170,119 @@ def test_g_z_symmetry():
...
+    coordinates = [easting, northing, upward]
+    # Compute g_northing gravity field on each computation point
+    results = point_mass_gravity(
+        coordinates, point_mass, masses, "g_northing", "cartesian"
+    )
+    npt.assert_allclose(results[0], -results[1])
+
+
+@pytest.mark.use_numba
+def test_g_northing_signal():
```
santisoler:
```suggestion
def test_g_northing_sign():
```

harmonica/tests/test_point_mass.py
```diff
@@ -91,6 +170,119 @@ def test_g_z_symmetry():
...
+    """
+    Test if g_northing field of a positive point mass has the correct signal
```
santisoler:
```suggestion
    Test if g_northing field of a positive point mass has the correct sign
```

harmonica/tests/test_point_mass.py
```diff
@@ -91,6 +170,119 @@ def test_g_z_symmetry():
...
+    """
+    # Define a single point mass
+    point_mass = [-10, 100.2, -300.7]
+    mass = [2670]
+    # Define three computation points located above the point mass, along the
+    # north axis
+    easting = np.zeros(3)
+    northing = np.array([0, 100.2, 210.7])
+    z = np.zeros(3)
```
santisoler:
```suggestion
    upward = np.zeros(3)
```

harmonica/tests/test_point_mass.py
```diff
@@ -91,6 +170,119 @@ def test_g_z_symmetry():
...
+    coordinates = [easting, northing, z]
```
santisoler:
```suggestion
    coordinates = [easting, northing, upward]
```

harmonica/tests/test_point_mass.py
```diff
@@ -91,6 +170,119 @@ def test_g_z_symmetry():
...
+    coordinates = [easting, northing, upward]
+    # Compute g_easting gravity field on each computation point
+    results = point_mass_gravity(
+        coordinates, point_mass, masses, "g_easting", "cartesian"
+    )
+    npt.assert_allclose(results[0], -results[1])
+
+
+@pytest.mark.use_numba
+def test_g_easting_signal():
```
santisoler:
```suggestion
def test_g_easting_sign():
```

harmonica/tests/test_point_mass.py
```diff
@@ -91,6 +170,119 @@ def test_g_z_symmetry():
...
+    """
+    Test if g_easting field of a positive point mass has the correct signal
```
santisoler:
```suggestion
    Test if g_easting field of a positive point mass has the correct sign
```

harmonica/tests/test_point_mass.py
```diff
@@ -91,6 +170,119 @@ def test_g_z_symmetry():
...
+    """
+    # Define a single point mass
+    point_mass = [-10, 100.2, -300.7]
+    mass = [2670]
+    # Define three computation points located above the point mass, along the
+    # east axis
+    easting = np.array([-150.7, -10, 79])
+    northing = np.zeros(3)
+    z = np.zeros(3)
```
santisoler:
```suggestion
    upward = np.zeros(3)
```

harmonica/tests/test_point_mass.py
```diff
@@ -91,6 +170,119 @@ def test_g_z_symmetry():
...
+    coordinates = [easting, northing, z]
```
santisoler:
```suggestion
    coordinates = [easting, northing, upward]
```

harmonica/forward/point_mass.py
birocoles: I just followed the same notation used in other functions of `point_mass.py`. There was no deprecation warning raise.

harmonica/forward/point_mass.py
birocoles: I don't know how to use LaTeX's here

santisoler: You are using LaTeX math mode actually. When using the `:math:`, Sphinx will interpret what's inside as LaTeX math. I was saying to use the `\text` macro like this:
```
:math:`g_\text{northing}(p)`, :math:`g_\text{easting}(p)` and :math:`g_\text{upward}(p)`
```

harmonica/forward/point_mass.py
santisoler: Ok. So, I don't know why passing some lists to Numba jitted functions seem to break things and with other list it has no trouble. I would change everything to tuples to prevent possible headaches in the future regarding lists. It seems that when Numba tries to work with lists (which are mutable), they need some special care because of their more complex structure when compared with tuples. What do you think?

birocoles: Ok for me @santisoler . I can change all list to tuples.

harmonica/forward/point_mass.py
birocoles: Ok! I can do this

harmonica/forward/point_mass.py
birocoles: I have replaced all lists by tuples to avoid problems with Numba

harmonica/forward/point_mass.py
birocoles: I have used `\text` for writing the subscripts of g_northing, g_easting and g_upward components

birocoles:
@santisoler , 

I have used the central finite difference formula for computing the approximated acceleration components in the tests `test_potential_versus_g_easting`, `test_potential_versus_g_northing` and `test_potential_versus_g_z`. In this case, the truncation error can be computed by using the classical formula for central finite difference. According to this formula, the truncation error is given by:

trunc_error

We could use this formula. What do you think?

santisoler:
That kind of error estimation is what I meant. Although I have my doubts about applying it as it is. We always struggle with how to tests analytical computations. On this case, computing the third derivative of our target field is even more complex than computing the field itself, so if we made a mistake on the computation of the field, it's very likely we would make the same mistake on the test function, resulting on a PASSED test although the error is hidden in there.

I'm thinking how to avoid computing this derivative. What about trying to bound the relative error instead of checking the result against the absolute error? For example, if we take the comparison of g_z against the potential, its relative error could be computed as:

imagen

If we assume that the epsilon is the same computation point as r, then the relative error is:

imagen

This expression for the relative error is not that complex, and we would only need to use the `distance_cartesian` function.

What do you think? I know there are a lot of simplifications on these calculations, but I don't think the test functions should be way more complex than the function we want to test.

birocoles:
Hi @santiago, it is ok for me. I will implement this today.

leouieda:
> Hi @santiago, it is ok for me. I will implement this today.

Nothing like some vacation time to get coding done :wink:

harmonica/forward/point_mass.py
leouieda: :+1: for this! We should think about why the tests didn't catch it as well.

leouieda:
@birocoles I think some of the changes here would be better suited for separate PRs. In particular, the spherical distance bug fix and replacing "gravity" with "gravitational" (each in a separate PR). That will help us keep the history clean and we can get those merged in quicker than this PR (which is getting big and taking a long time).

harmonica/forward/point_mass.py
santisoler: You are right @leouieda. Tests weren't failing because we only test potential symmetry and not the true value of the potential on a certain point. The potential holds the same symmetry both in Cartesian and spherical coordinates: two observation points located along the longitudinal axis at the same distance of the point mass measure the same gravity potential, and the same happens with two observation points located along the easting axis at the same distance of the point mass.

birocoles:
ok @leouida, no problems for me. But I don’t know how to split this PR 😅

leouieda:
@birocoles that depends on how well you curated you git commits :) I'll see if I can separate them without much work. Otherwise, we can keep this as is. It's a lot easier to separate things from the start.

leouieda:
Right, the changes are all mixed together in the same commits. So there is no way to separate renaming things to gravitational and the bug fix from the other changes. This is one reason to remember to commit often and keep changes separate for commits. We're all guilty of doing this but in situations like this is when we're screwed for not having it.

Another thing you could do is create a new branch and manually add the bug fix (using the right distance function). We could do it for you but I want you to have credit for having found it and fixed it.

birocoles:
I will split things into three parts: 1) gravity to gravitational, 2) function distance and 3) all the remaining stuff. Ok?

leouieda:
@birocoles I'm merging in the latest changes from #116 and #117 here and resolving the conflicts.

leouieda:
Done! Thanks for all your help with this! What is left to do here? Are we just waiting for the finite difference tests? If so, I'd suggest merging this in and implementing those in a separate PR as they will require some experimentation still (and this PR is getting big and out of hand).

birocoles:
Sorry @leouieda , I got confused here. I have created another branch with the new functions for computing g_northing and g_easting, as well as the related tests and documentation modifications. I thought you would shut this PR down. I have just reviewed the new branch for creating a new PR. Do you prefer to take a look in the new PR or preceed with this one?

leouieda:
@birocoles oh sorry for the confusion! Its up to you want to use this one or create a new one. There shouldn't be any difference. Did you make any changes there that aren't included in this branch?

birocoles:
I think the things are more organized there. I prefer to create a new PR and close the this one.

leouieda:
:+1: whatever is best for you. Closing this in favor of #119

PR #98, opened 2019-09-09
Add sanity check for the size of masses array

### LINKED ISSUE #97: Raise error if masses array has not the same number of elements as points

On the point mass forward model would be nice to check if the `masses` has the same number of elements as the points coordinates inside `points`. On the tesseroid forward model this is done, but for some reason we forgot to add it on the point mass case.

santisoler:
Check if the masses array has the same size as the point masses
coordinates and raise an error if it doesn't. Add test function for the
new sanity check.

Fixes #97

**Reminders**


harmonica/forward/point_mass.py
```diff
@@ -178,9 +178,19 @@ def point_mass_gravity(
...
     result = np.zeros(cast.size, dtype=dtype)
     # Prepare arrays to be passed to the jitted functions
-    coordinates = (np.atleast_1d(i).ravel().astype(dtype) for i in coordinates[:3])
-    points = (np.atleast_1d(i).ravel().astype(dtype) for i in points[:3])
+    coordinates = np.vstack(
+        tuple(np.atleast_1d(i).ravel().astype(dtype) for i in coordinates[:3])
+    )
+    points = np.vstack(
+        tuple(np.atleast_1d(i).ravel().astype(dtype) for i in points[:3])
+    )
```
leouieda: @santisoler I'm not sure why this is needed. Did you come across an error? Numba should be fine with tuples since we don't ever operate on the entire `coordinates`, only one array at a time. It's worth checking is `np.vstack` isn't re-allocating the memory for the arrays. If it is, then this is definitely something we should avoid.

santisoler: I don't know what happened when I wrote this. Stacking the coordinates and points is completely unnecessary, so I went back to defining them as tuples.

harmonica/forward/point_mass.py
```diff
@@ -178,9 +178,14 @@ def point_mass_gravity(
...
     # Prepare arrays to be passed to the jitted functions
-    coordinates = (np.atleast_1d(i).ravel().astype(dtype) for i in coordinates[:3])
-    points = (np.atleast_1d(i).ravel().astype(dtype) for i in points[:3])
+    coordinates = tuple(np.atleast_1d(i).ravel().astype(dtype) for i in coordinates[:3])
+    points = tuple(np.atleast_1d(i).ravel().astype(dtype) for i in points[:3])
     masses = np.atleast_1d(masses).astype(dtype).ravel()
+    # Sanity checks
+    if masses.size != points[0].size:
+        raise ValueError(
+            "Masses array must have the same size as number of point masses."
```
leouieda: It would be good to include the values in the error message so we know what's wrong. This message could be triggered deep in a function call stack and the user might have no idea what is going on.

santisoler: Done! Let me know if you like it or I should change anything. Besides, I've opened #104 in order to improve the tesseroids error message for invalid density array.

leouieda:
:+1: merging!

PR #104, opened 2019-09-20
Improve error message for invalid density array

santisoler:
Add size of density array and number of tesseroids on the error message
when they mismatch in order to make clearer what's failing.

The improvement of such error messages was triggered after #98.

**Reminders**


leouieda:
Nice :beers: Thanks, Santi

Issue #93, opened 2019-08-19
Speed up EQL for high number of data points


santisoler:
**Description of the desired feature**

Would be nice to improve the harmonic Equivalent Layer gridder in order to speed up its computations when working with a great number of data points.
Firstly, would be nice to make the Jacobian matrix sparse. This could be done by setting to zero those elements that involve distant data and source points.

Further techniques could be related to reducing the computation time involved on the forward modelling, reducing the number of point sources, etc. These alternatives could be investigated separately, so we could focus on the first one.

**Are you willing to help implement and maintain this feature?** Yes

santisoler:
This will be solved with the implementation of Gradient Boosted Equivalent Layer. Closing this.

PR #106, opened 2019-10-01
WIP Add option to build a sparse Jacobian matrix on EQL

santisoler:
Make EQLHarmonic capable of building the Jacobian matrix as a sparse matrix in order to reduce memory consumption. The elements of the Jacobian that involve distant source and computation points are neglected and assumed to be equal to zero. The elements are neglected if the distance between these two points is larger than a cutoff distance that is passed as argument to the gridder constructor.

Partially solves #93 

**Reminders**


santisoler:
I'll close this issue. We haven't got any positive outcome by trying sparse Jacobian matrices for EQLs.

PR #109, opened 2019-10-02
Avoid unnecessary copy of input arrays in forward models

### LINKED ISSUE #100: Avoid stacking coordinates on tesseroid forward modelling

As raised by @leouieda on #98 and #86, we should avoid stacking `coordiantes` arrays in order to avoid unnecessary memory allocation. Therefore, `coordinates` must be a tuple of 1d numpy arrays. This issue was already solved on point mass and prisms forward models, but the tesseroids forward model still stacks `coordinates`.

### LINKED ISSUE #107: Avoid casting data type on forward models

The current forward models (prisms, point masses and tesseroids) cast the data type of some (or all) input arrays. The original idea behind it was to avoid working with input arrays full of integers (that could produce wrong results or even Numba errors). Nevertheless, casting the type of the inputs arrays implies copying them, resulting in an unnecessary duplication of memory.
We should stop casting the type of the input arrays on every forward model, and probably raise a warning (or an error) in case the passed arrays are full of integers.

This issue was raised after #100.

santisoler:
Avoid casting the data type of input arrays in forward models. The dtype
argument is used only for building the arrays where the results are
being stored. Make _longitude_continuity to copy the tesseroids array in
order to avoid modifying the original tesseroids array passed by the
user. Avoid stacking the coordinates on the tesseroid forward model.
Improve how invalid tesseroids are caught and how longitude continuity
is applied.

Fixes #100 fixes #107

**Reminders**


santisoler:
@leuieda I have a small doubt regarding the following line:
https://github.com/fatiando/harmonica/blob/50977376873febfa41e4d9743217b3faa763442a/harmonica/forward/tesseroid.py#L257-L258

Because the coordinates passed to `jit_tesseroid_gravity()` is a tuple, I had to modify how the coordinates of the computation point on each iteration is passed to the `_adaptive_discretization()` and `jit_point_mass_spherical()` functions.
What do you think about passing the computation point to `_adaptive_discretization()` like this? Is it alright or is that line allocating memory?

harmonica/forward/tesseroid.py
```diff
@@ -157,10 +154,10 @@ def tesseroid_gravity(
...
     n_nodes, glq_nodes, glq_weights = glq_nodes_weights(glq_degrees)
     # Initialize arrays to perform memory allocation only once
-    stack = np.empty((stack_size, 6), dtype=dtype)
-    small_tesseroids = np.empty((max_discretizations, 6), dtype=dtype)
-    point_masses = np.empty((3, n_nodes * max_discretizations), dtype=dtype)
-    weights = np.empty(n_nodes * max_discretizations, dtype=dtype)
+    stack = np.empty((stack_size, 6), dtype=tesseroids.dtype)
+    small_tesseroids = np.empty((max_discretizations, 6), dtype=tesseroids.dtype)
+    point_masses = np.empty((3, n_nodes * max_discretizations), dtype=tesseroids.dtype)
+    weights = np.empty(n_nodes * max_discretizations, dtype=density.dtype)
```
leouieda: @santisoler any reason to not use the given `dtype` here?

harmonica/forward/tesseroid.py
```diff
@@ -658,10 +653,18 @@ def _longitude_continuity(tesseroids):
...
+    -------
+    tesseroids : 2d-array
+        Modified boundaries of the tesseroids.
     """
+    # Copy the tesseroids in order to avoid modifying the original tesseroids array
+    tesseroids = tesseroids.copy()
     west, east = tuple(tesseroids[:, i] for i in range(2))
     west %= 360
     east %= 360
     tess_to_be_changed = west > east
```
leouieda: I would put these two lines below and the copy in a `if np.any(tess_to_be_changed)` block to avoid copying if we don't need to.

leouieda:
> What do you think about passing the computation point to `_adaptive_discretization()` like this? Is it alright or is that line allocating memory?

It should be fine since tuples only hold the reference to the numbers.

harmonica/forward/tesseroid.py
santisoler: No special reason, but I think we should use the given `dtype`, so if the user wants to reduce memory consumption by passing a `np.float32`, each one of these array will occupy less memory.

PS: Multiple line comments! :muscle:

harmonica/forward/tesseroid.py
santisoler: You're absolutely right!

santisoler: **Update**: I don't think we can copy only the west and east boundaries, because when we need to incorporate the modified east and west into the tesseroids array, a completely new tesseroids array must be created.

harmonica/forward/tesseroid.py
leouieda: Right, so that means we should be passing `dtype=dtype` instead of `dtype=tesseroids.dtype`, no?

harmonica/forward/tesseroid.py
leouieda: >  I don't think we can copy only the west and east boundaries

I agree. But if `np.any(tess_to_be_changed) == False` then we don't need to copy `tesseroids` at all and could return the original tesseroids instead. Right?

harmonica/forward/tesseroid.py
```diff
@@ -658,10 +653,18 @@ def _longitude_continuity(tesseroids):
...
         All tesseroids must have valid boundary coordinates.
+
+    Returns
+    -------
+    tesseroids : 2d-array
+        Modified boundaries of the tesseroids.
     """
+    # Copy the tesseroids in order to avoid modifying the original tesseroids array
+    tesseroids = tesseroids.copy()
     west, east = tuple(tesseroids[:, i] for i in range(2))
```
leouieda:
```suggestion
    west, east = tesseroids[:, 0], tesseroids[:, 1]
```

harmonica/forward/tesseroid.py
leouieda: In that sense, do lines 665 and 666 (:metal:) need to be inplace operations?

harmonica/forward/tesseroid.py
santisoler: Yes! I've already commited that. Now those lines have the original `dtype=dtype`.

harmonica/forward/tesseroid.py
santisoler: Each time I go back with this longitude continuity functions I have to rethink everything (I'm thinking to write a little document to explain it to myself).

Lines 665 and 666 move every longitude boundary to the [0, 360) interval. So, if we passed a tesseroid like (350, 370, ...), it will move it to (350, 10, ...). Then the `tess_to_be_changed` will  mark it as needed to be changed, moving it to the [-180, 180) interval: (-10, 10, ...).

So, lines 665 and 666 must be run independently if no tesseroid should be moved to [-180, 180) afterwards.

Maybe we can check if `west` and `east` fall either on [0, 360) or [-180, 180) and if `west < east`. If any of these fail, then we should perform the transformation of the tesseroids, i.e. copying it. That's the only option I can think to avoid copying the entire tesseroids array every time we run the forward model independently of the boundaries we pass.

harmonica/forward/tesseroid.py
leouieda: Oops, missed that :disappointed:

harmonica/forward/tesseroid.py
leouieda: What if we run `_check_tesserroids` first and assume that all are either [0, 360] or [-180, 180]? That seems reasonable to me. Allowing `(350, 370, ...)` feels wrong. If that's the case, then we can avoid the in place `%` operation (and the copy it requires).

santisoler: Ok, I've updated the standards for what we consider a valid tesseroid.

A valid tesseroid should have:
- latitudinal boundaries within the [-90, 90] degrees interval,
- north boundaries greater or equal than the south boundaries,
- radial boundaries positive or zero,
- top boundaries greater or equal than the bottom boundaries,
- longitudinal boundaries within the [-180, 360] degrees interval,
- longitudinal interval must not be greater than one turn around the globe.

Some valid tesseroids have its west boundary greater than the east one, e.g.
``(350, 10, ...)``. On these cases the ``_longitude_continuity`` function is applied
in order to move the longitudinal coordinates to the [-180, 180) interval. Any valid
tesseroid should have east boundaries greater than the west boundaries before or
after applying longitude continuity.

leouieda: :+1: looks good to me! Should the `_longitude_continuity` function be deleted now?

santisoler: Nope. I call it from the `_check_tesseroids` function. I decided to leave it on a separate function because `_check_tesseroids` is already very cluttered.

santisoler:
I'm thinking to add a `test_valid_tesseroid` function along with the `test_invalid_tesseroid` in order to test if no valid tesseroid is caught as invalid.

harmonica/forward/tesseroid.py
leouieda: :+1: right, I missed that. Carry on :)

santisoler:
@leouieda I think this is ready to be merged. Would you add something else?

leouieda:
👍 good to go for me. Just need to fix the conflicts after #110

leouieda:
Conflicts fixed. Merging this :rocket:

PR #110, opened 2019-10-03
Allow avoiding sanity checks on forward models

### LINKED ISSUE #108: Add option to avoid sanity checks on forward models

The current forward models for [prisms](https://www.fatiando.org/harmonica/dev/api/generated/harmonica.prism_gravity.html#harmonica.prism_gravity) and [tesseroids](https://www.fatiando.org/harmonica/dev/api/generated/harmonica.tesseroid_gravity.html#harmonica.tesseroid_gravity) perform sanity checks on the input arrays. Most of them involve looping through computation points and sources, resulting in a task that could take a significant amount of time. Although this time is negligible when performing a single forward computation, it could be problematic when several forward computations are needed, e.g. on inversions. Would be nice to include an optional argument to avoid performing these sanity checks and also add a notice to only disable them when it's really needed and in a very careful way.

A cool name for the new optional argument could be `disable_checks` that should be False by default.
The sanity checks that should be optionally disabled are the one that make use of a private function, like [_check_prisms()](https://github.com/fatiando/harmonica/blob/master/harmonica/forward/prism.py#L103).

The new optional argument should have a description on the docstring, explaining when you would set it to True and its possible unwanted consequences.

ziebam:
Adds the `disable_checks` flag to the forward models of prisms and tesseroids, as proposed in #108. The solution is up for discussion, as it is my first contribution to an actual codebase. I ran into some issues when trying to test the code after commiting the changes and I would appreciate help regarding the problem so I can add the appropriate tests.

The error is as follows:
```
============================= test session starts =============================
platform win32 -- Python 3.7.3, pytest-5.2.0, py-1.8.0, pluggy-0.13.0 -- D:\Anaconda3\envs\harmonica\python.exe
cachedir: .pytest_cache
rootdir: D:\dev\python\harmonica, inifile: pytest.ini
plugins: cov-2.7.1
collecting ... collected 0 items

============================ no tests ran in 0.00s ============================
ERROR: file or package not found: harmonica (missing __init__.py?)
```

I also don't want to mess with the documentation
without permission. Is it okay if I cover the new flag accordingly?

Fixes #108

**Reminders**

- [X] Run `make format` and `make check` to make sure the code follows the style guide.
- [X] Add new public functions/methods/classes to `doc/api/index.rst`.
- [X] Write detailed docstrings for all functions/methods.


santisoler:
Nice work @Pralkarz! This PR looks promising!

Regarding the error you're getting when running the tests, have you installed Harmonica? It seems it's not installed on your `harmonica` conda environment. Try to install it through the `Makefile` (remember to change to the `harmonica` environment first):
```
make install
```

> I also don't want to mess with the documentation
> without permission. Is it okay if I cover the new flag accordingly?

Feel free to edit the documentation as you please. Any PR is reviewed before being merged, so apply any commit you think it's worthy and then we can discuss it, or just open the question and we can discuss it before commiting it. Just remember that each PR must solve a specific problem, so if you are thinking to enhance something that is not related to the subject of this PR, you can always open an issue to discuss it, or just open a PR implementing it.

One of the things I haven't mentioned on the issue is that any PR must meet a 100% test coverage in order to be ready to be merged. It's a very strict standard to work with, but we found from experience that once an untested code is merged there's a very little change that the tests will be added in the future.
Although you are not going to implement a big enhance, that `if` statement will create an untested case: what happens when `disable_coords` is True. Do not worry too much about this right now, just finish improving the documentation as you wish and then we can discuss how to add simple tests for the new case.

Pralkarz:
I haven't worked with Sphinx before, so I didn't know it added the `disable_checks` documentation automatically. That's what I wanted to include. For now I "fixed" the formatting by disabling one of the `pylint` warnings, but I'm open for suggestions on how to implement that better. Please see https://github.com/fatiando/harmonica/pull/110/commits/28dd57755daf95bd4569bb3fc65bedd826fead1c for reference.

If there aren't any at the moment, I think I could use some guidance in writing the test. I can't think of a good way to check if the given part of code hasn't been run.

santisoler:
> I haven't worked with Sphinx before, so I didn't know it added the `disable_checks` documentation automatically. That's what I wanted to include.

Luckily we use the sphinx-autogen plugin that automatically builds the API documentation by reading the source files. We just need to remember to add an entry for any new module on `doc/api/index.rst`. But in this case that's not needed because we already have included the forward models to the index.
You can build the documentation and check how it's rendering through:
```
make -C doc; make -C serve
```
For a fresh build of the docs, run the following before building it:
```
make -C doc clean
```

> For now I only "fixed" the error that was causing the style check to fail by disabling one of the `pylint` warnings, but I'm open for suggestions on how to implement that better. Please see [28dd577](https://github.com/fatiando/harmonica/commit/28dd57755daf95bd4569bb3fc65bedd826fead1c) for reference.

That's right. We try to follow the pep8 guidelines, but on some cases we decide to violate them for any specific need. I think this case falls under this category.

> If there aren't any, I think I could use some guidance in writing the test. I can't think of a good way to check if the given part of code hasn't been run.

Writing test functions is tricky. 
One of the tests that comes to my head is to compute a simple forward model passing a legit source (for example a prism with correct boundaries) enabling and disabling the sanity checks. Both forward models should return the same result without raising any error.

Another possible test is to build a non legit source (prisms with wrong boundaries, e.g. `[0, 10, 0, 10, 0, -10]`), and check if when computing the forward model with `disable_checks` to True, it doesn't raise an error. This could be included inside the already existing [`test_invalid_prisms()`](https://github.com/fatiando/harmonica/blob/master/harmonica/tests/test_prism.py#L37) function.

For this case I think we can compute a simple forward model passing a legit source (for example a prism with correct boundaries) enabling the sanity checks, and then we can pass compute the same forward model disabling the checks, but now we exchange the top and bottom boundaries (for example).

All test functions could be included inside [`harmonica/tests/test_prisms.py`](https://github.com/fatiando/harmonica/blob/master/harmonica/tests/test_prism.py) and [`harmonica/tests/test_tesseroid.py`](https://github.com/fatiando/harmonica/blob/master/harmonica/tests/test_tesseroid.py).

Pralkarz:
I suggest to simply check if the code doesn't raise any errors for invalid data when the flag is set to `True`. I don't think checking valid data with the flag ticked on/off is necessary, as it should only be used when we're certain of its rationality anyway.

I'll get to writing these tests tomorrow, as it's pretty late where I live right now, however I will also need some guidance in specifying an invalid tesseroid. To be honest I decided to contribute to Harmonica as the issue seemed very simple (and it was very simple indeed), but I lack the scientific background to be able to understand the math and physics behind prisms and tesseroids. Could you provide me with a sample invalid tesseroid or point me to a resource which could help me dive in the topic? The repository seems very well-organized and I would like to contribute more in the future, so it would be a good idea to learn something about that stuff.

santisoler:
> I suggest to simply check if the code doesn't raise any errors for invalid data when the flag is set to `True`.

I completely agree!

> I don't think checking valid data with the flag ticked on/off is necessary, as it should only be used when we're certain of its rationality anyway.

But I disagree on this one. All current tests for tesseroids and prisms are run with the sanity checks enabled, what is the best way to proceed on the general case. Nevertheless I think we should include at least one simple test to check what happens when the sanity checks are not run. On the prisms case it's kind of trivial (because the sanity check only checks if the boundaries of the prism are well defined), but the tesseroid sanity checks are far more complex (in fact the `_longitude_continuity()` function modifies the boundaries of the tesseroids).


> I'll get to writing these tests tomorrow, as it's pretty late where I live right now, however I will also need some guidance in specifying an invalid tesseroid.

No problem! There's no hurry to finish this, so take your time.

> To be honest I decided to contribute to Harmonica as the issue seemed very simple (and it was very simple indeed), but I lack the scientific background to be able to understand the math and physics behind prisms and tesseroids. Could you provide me with a sample invalid tesseroid or point me to a resource which could help me dive in the topic?

Of course, we are very used to work with these forward models and have a lot of things in mind that most people would obviously ignore if they don't work with them.

These two forward models (prisms and tesseroids) seem very alike, but on the background they are significantly different. Firstly, the prisms (rectangular prisms to be more precise) are geometrical bodies defined on a Cartesian coordinate system. They are broadly used to model underground structures among geosciences due to the fact that the gravitational field they generate have analytical solution, i.e. we just need to evaluate a certain function to get the value of the gravitational field. This seems an almost perfect tool to work with, but it has a drawback: when trying to model big regional or global areas, it's not possible for prisms to account for the curvature of the Earth (you can rotate each prism, but that involve more computation and some numerical errors due to overlap will be included).
Here's where tesseroids are very handy. Also know as "spherical prisms", we define them in spherical coordinates by 3 boundaries: two radial, two latitudinal and two longitudinal boundaries. They do account for the curvature of the Earth, but there's no analytical solution for the gravitational fields they create, therefore we need to perform numerical computations to approximate it.

In summary: when working with small local areas, using prisms is the way to go. But when working with regional or global areas (an entire continent for example), you better use tesseroids.

If you want to know more about these two forward models I can point you to the following papers:

**For prisms:**
- Nagy (2002) Nagy, D., Papp, G. & Benedek, J.(2000). The gravitational potential and its derivatives for the prism. Journal of Geodesy 74: 552. doi:[10.1007/s001900000116](https://doi.org/10.1007)/s001900000116
- Fukushima, T. (2019). Fast computation of prismatic gravitational field. doi:[10.13140/RG.2.2.30598.93766](https://doi.org/10.13140/RG.2.2.30598.93766)

**For tesseroids:**
- Uieda, L., V. Barbosa, and C. Braitenberg (2016), Tesseroids: Forward-modeling gravitational fields in spherical coordinates, GEOPHYSICS, F41–F48, doi:[10.1190/geo2015-0204.1](https://doi.org/10.1190/geo2015-0204.1)
- Soler, S. R., Pesce, A., Gimenez, M. E., & Uieda, L., 2019. Gravitational field calculation in spherical coordinates using variable densities in depth, Geophysical Journal International, doi:[10.1093/gji/ggz277](https://doi.org/10.1093/gji/ggz277)


When passing a tesseroid or a prism to our forward model functions, we pass their boundaries in a specific order. For example, each prism must be passed as `(west, east, south, north, bottom, top)`, where `west` and `east` are its easting boundaries, `south` and `north` its northing boundaries and `bottom` and `top` its upward boundaries. A valid prism must not have a `west` greater than `east`, nor `south` greater than `north` nor `top` than `bottom`. So an invalid prism would be: `(0, 10, 0, 10, 0, -10)` because its `top` is lower than the `bottom`.

Tesseroids are passed in a similar way:  `(west, east, south, north, bottom, top)`, but now the `west`, `east`, `south` and `north` boundaries are degrees, and the `bottom` and `top` are radial distances. Àny valid tesseroid should not have `south` greater than `north` nor `bottom` greater than `top`. But it could have a `west` *numerically* greater than `east` (for example `west=350` and `east=10` it's completely valid, it's a tesseroid with a longitudinal size of 20 degrees, equivalent to `west=-10` and `east=10`). That's why we use the `_longitude_continuity()` function to move the longitudinal boundaries of each tesseroid to an equivalent one that respects that `west <= east`. We need to do this because on some point we compute the difference between the two longitudinal boundaries, therefore they cannot be phase shifted.
Finally, the tesseroid forward model has a limitation that the prism one doesn't have: the computation point must not be inside the tesseroid, that's why we check if any passed computation point is inside any tesseroid.

> The repository seems very well-organized and I would like to contribute more in the future, so it would be a good idea to learn something about that stuff.

That would be awesome! We are glad to get help! :clap:

harmonica/forward/prism.py
```diff
@@ -51,6 +53,10 @@ def prism_gravity(coordinates, prisms, density, field, dtype="float64"):
     dtype : data-type (optional)
         Data type assigned to prism boundaries, computation points coordinates and
         resulting gravitational field. Default to ``np.float64``.
+    disable_checks : bool (optional)
+        Flag that controls whether to perform a sanity check on the model. Should be set
+        to ``True`` when it is certain that the material is rational and it does not need
```
leouieda:
```suggestion
        to ``True`` only when it is certain that the input model is valid and it does not need
```

harmonica/forward/tesseroid.py
```diff
@@ -95,6 +96,10 @@ def tesseroid_gravity(
     dtype : data-type (optional)
         Data type assigned to tesseroid boundaries, computation points coordinates and
         resulting gravitational field. Default to ``np.float64``.
+    disable_checks : bool (optional)
+        Flag that controls whether to perform a sanity check on the model. Should be set
+        to ``True`` when it is certain that the material is rational and it does not need
+        to be checked. Default to ``False``.
```
leouieda: Same suggestions as above.

leouieda:
@Pralkarz thank you for submitting this PR! :tada: As @santisoler said, we're really glad to get some help. Left a few suggestions for the wording on the docstring. Seems to me like this is almost done and only missing a test case.

One suggestions would be to create a prism/tesseroid with top and bottom boundaries flipped around. In this case, if checks are off you should get the same output only with the sign inverted (for example, the correct value would be `1.5` and you'll get `-1.5`.

Pralkarz:
@leouieda Thanks for your suggestions, I improved the wording as per your request. I'm not a native English speaker, so that's definitely appreciated.

I also included the requested tests for the disable_checks flag in the form of two similar functions (`test_disable_checks`) in `test_prism.py` and `test_tesseroid.py`. Although they adhere to `black`, `flake8` and `pylint` standard, they look a bit sloppy, so if you have an idea on how to improve them – please let me know! For now I'm marking this PR as ready for review as all checks have passed and I've successfully implemented the requested feature.

harmonica/tests/test_prism.py
```diff
@@ -52,6 +52,23 @@ def test_invalid_prisms():
         _check_prisms(np.atleast_2d([w, e, s, n, top, bottom]))
 
 
+def test_disable_checks():
```
santisoler:
```suggestion
@pytest.mark.use_numba
def test_disable_checks():
```

harmonica/tests/test_prism.py
```diff
@@ -52,6 +52,23 @@ def test_invalid_prisms():
...
+    coordinates = [0, 0, 0]
+    # Check if an invalid prism doesn't raise an error with the disable_checks flag set to True
+    prism_gravity(
+        coordinates, invalid_prism, density, field="potential", disable_checks=True
+    )
+    # Check if an invalid prism returns the inverse of a valid one's return value
+    # with only top and bottom values flipped between the two.
+    assert prism_gravity(
+        coordinates, invalid_prism, density, field="potential", disable_checks=True
+    ) == (-1) * prism_gravity(coordinates, valid_prism, density, field="potential")
```
santisoler: This checks are great! I would simplify them in order to reduce repeated code and to avoid computing the same quantity twice: the gravity potential generated using the invalid prism is computed twice on lines 62 and 67.
One solution could be to store this result:
```python
    # Check if an invalid prism doesn't raise an error with the disable_checks flag set to True
    invalid_result = prism_gravity(
        coordinates, invalid_prism, density, field="potential", disable_checks=True
    )
    # Check if the invalid prism generates a potential field inverse
    # than the one generated by the valid prism.
    valid_result = prism_gravity(coordinates, valid_prism, density, field="potential")
    assert invalid_result == - valid_result
```

harmonica/tests/test_tesseroid.py
```diff
@@ -159,6 +159,25 @@ def test_invalid_tesseroids():
             _check_tesseroids(new_tesseroids)
 
 
+def test_disable_checks():
```
santisoler:
```suggestion
@pytest.mark.use_numba
def test_disable_checks():
```

harmonica/tests/test_tesseroid.py
```diff
@@ -159,6 +159,25 @@ def test_invalid_tesseroids():
...
+    tesseroid_gravity(
+        coordinates, invalid_tesseroid, density, field="potential", disable_checks=True
+    )
+    # Check if an invalid tesseroid returns the inverse of a valid one's return value
+    # with only top and bottom values flipped between the two.
+    assert tesseroid_gravity(
+        coordinates, invalid_tesseroid, density, field="potential", disable_checks=True
+    ) == (-1) * tesseroid_gravity(
+        coordinates, valid_tesseroid, density, field="potential"
+    )
```
santisoler: I would change it in a similar way I suggested for the prism case: store the resulting potential generated by passing the invalid tesseroid.

ziebam:
I've simplified the tests according to your review. Thanks for the suggestion! I thought about storing the results in variables at first, but for some reason I thought it would be less memory-efficient than the previous approach.

As to `assert invalid_result == -valid_result` lines, when formatting the code with `make format` `black` was automatically removing the whitespace between `-` and `valid_result`, but only in `test_tesseroid.py` for some reason, so I removed the space in `test_prism.py` as well.

santisoler:
Thanks @ziebam ! I think this PR is ready to be merged. I can do so if you agree.

By the way, I just noticed you made all these commits on the master branch of your fork. It's always better to create a new branch on your fork in order to open a PR. The idea is to keep the master branch of our forks to be updated with the master branch from the original repo. Don't worry for this now, the PR can still be merged without any problem. I recommend you to delete the fork after it's merged (and fork Harmonica again whenever you want to submit a new PR), or reset the master branch to the last common commit between `harmonica:master` and `ziebam:master`, but this is more complicated.

harmonica/tests/test_prism.py
```diff
@@ -52,6 +52,23 @@ def test_invalid_prisms():
...
+    density = 100
+    coordinates = [0, 0, 0]
+    # Check if an invalid prism doesn't raise an error with the disable_checks flag set to True
+    invalid_result = prism_gravity(
+        coordinates, invalid_prism, density, field="potential", disable_checks=True
+    )
+    # Check if the invalid prism generates a potential field inverse
+    # than the one generated by the valid prism
+    valid_result = prism_gravity(coordinates, valid_prism, density, field="potential")
+    assert invalid_result == -valid_result
```
leouieda:
```suggestion
    npt.assert_allclose(invalid_result, -valid_result)
```

When comparing floating point numbers, it's always better to use `numpy.testing.assert_allclose` or `numpy.testing.assert_almost_equal` because they let you set a desired precision. Comparing with `==` tends to fail on different OSs or when packages update (I've suffered from this too many times in the past :-)

harmonica/tests/test_prism.py
```diff
@@ -52,6 +52,23 @@ def test_invalid_prisms():
         _check_prisms(np.atleast_2d([w, e, s, n, top, bottom]))
 
 
+@pytest.mark.use_numba
+def test_disable_checks():
+    "Check if the disable_checks flag works properly"
+    valid_prism = [0, 10, 0, 10, -10, 0]
+    invalid_prism = [0, 10, 0, 10, 0, -10]
+    density = 100
+    coordinates = [0, 0, 0]
```
leouieda:
```suggestion
    coordinates = [0, 0, 0]
    # By default, an error should be raised for invalid input
    with pytest.raises(ValueError):
        prism_gravity(coordinates, invalid_prism, density, field="potential")
```

Added a test that the error is still raised by default. Just to complement the existing tests.

leouieda:
CIs seem to be happy with the changes so I'm going to go ahead and merge this. Thank you @ziebam :clap: :tada: I'm glad you were able to learn something from this! We really appreciate the help/ We have been transitioning to a [Slack group](https://join.slack.com/t/fatiando/shared_invite/enQtNzY4NDQ3ODQwNDk4LTc5MTU5OWNkNTczMDY4NjcyNjcyZTU0Y2I3MDQ0NWUwMmEzMTBkNmVjNTExMGZkMjA0YzM1OGYyMzZlMDk3YTU) for discussion so feel free to join that if you want to get more involved in the project.

PR #113, opened 2019-10-11
Make gz effects more obvious in point mass example

leouieda:
The previous version was using viridis as a colormap with sources
generating very smooth fields due to their depth. I made the sources a
bit shallower so they have more pronounced lobes and used seismic to
bring out the positive and negative effects.

**Reminders**

- [X] Run `make format` and `make check` to make sure the code follows the style guide.
- [X] Add tests for new features or tests that would have caught the bug that you're fixing.
- [X] Add new public functions/methods/classes to `doc/api/index.rst`.
- [X] Write detailed docstrings for all functions/methods.
- [X] If adding new functionality, add an example to the docstring, gallery, and/or tutorials.


leouieda:
This is the new plot for the example:
sphx_glr_point_mass_001

leouieda:
Actually, this might be a better plot:
sphx_glr_point_mass_001

leouieda:
Yes, that is usually true. For this case, the grid resolution doesn’t really matter because it isn’t real data. That’s why I chose the contours.

leouieda:
Merging this in 👍

PR #116, opened 2019-10-23
Distance function in kernel_potential_cartesian

birocoles:
I have replaced `distance_spherical` by `distance_cartesian` in the function `kernel_potential_cartesian`. I have also removed the unused `distance_spherical` function from `point_mass_gravity.py`. The functions computing things in spherical coordinates use the function `distance_spherical_core`.

**Reminders**


leouieda:
@santisoler @birocoles can you think of a test that would have caught this bug?  We'd need something that checks the absolute values. One idea would be to generate some outputs from `fatiando` and tests against those.

birocoles:
Hi @leouieda and @santisoler,

I can create a test comparing the output with reference values computed independently. What do you think?

leouieda:
@birocoles that would be good, thanks! It can serve as a regression test (checking that results continue to be the same).

birocoles:
Hi @leouieda and @santisoler , I have computed, by using an external code, some reference values for the gravitational potential produced by a point mass. These values were used to create the `test_potential_cartesian_known_values` in `test_point_mass.py`. I have verified that this test fails if the `distance_spherical` function is used in `kernel_potential_cartesian` instead of the function `distance_cartesian`.

santisoler:
Thanks @birocoles for splitting the original PR! I know it's hard work and it's only triggered by maintaining good practices, but in the end they save us time and a few headaches.

PR #120, opened 2019-11-06
Add ground and airborne synthetic surveys

santisoler:
Add functions to create synthetic ground and airborne surveys based on the
gravity dataset from South Africa and the total-field magnetic anomaly dataset
from Great Britain. The synthetic surveys are samples from the original
datasets that are rescaled to the desired region defined in geodetic
coordinates.

**Reminders**


santisoler:
@leouieda I'm thinking I can extend the functionalities of these two functions in order to return the synthetic surveys on either Cartesian or geodetic coordinate systems. It wouldn't be hard to implement because the original surveys are already given in geodetic. I'm thinking to add a `coordinate_system` argument to each function through which the user can change the desired coordinate system. What do you think?

harmonica/synthetic/surveys.py
```diff
@@ -0,0 +1,145 @@
...
+    Returns
+    -------
+    survey : :class:`pandas.DataFrame`
+        Dataframe containing the coordinates of the observation points and their
+        elevation Cartesian coordinates for the synthetic model. All coordinates and
+        altitude are in meters.
+    """
+    # Cut the data into the cut_region
+    inside_points = inside((survey.longitude, survey.latitude), cut_region)
+    survey = survey[inside_points].copy()
```
santisoler: By the way, I'm struggling with a pandas warning if I don't copy the `survey` DataFrame.
If I remove the `.copy` I get the following warning:

```traceback
/home/santi/.miniconda3/envs/harmonica/lib/python3.7/site-packages/numba/types/containers.py:3
  /home/santi/.miniconda3/envs/harmonica/lib/python3.7/site-packages/numba/types/containers.py:3: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    from collections import Iterable

harmonica/tests/test_synthetic_surveys.py::test_ground_survey
harmonica/tests/test_synthetic_surveys.py::test_airborne_survey
harmonica/tests/test_synthetic_surveys.py::test_cut_region_ground_survey
harmonica/tests/test_synthetic_surveys.py::test_cut_region_airborne_survey
  /home/santi/git/harmonica/harmonica/synthetic/surveys.py:130: SettingWithCopyWarning: 
  A value is trying to be set on a copy of a slice from a DataFrame.
  Try using .loc[row_indexer,col_indexer] = value instead
  
  See the caveats in the documentation: http://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
    survey.longitude.values, survey.latitude.values

harmonica/tests/test_synthetic_surveys.py::test_ground_survey
harmonica/tests/test_synthetic_surveys.py::test_airborne_survey
harmonica/tests/test_synthetic_surveys.py::test_cut_region_ground_survey
harmonica/tests/test_synthetic_surveys.py::test_cut_region_airborne_survey
  /home/santi/git/harmonica/harmonica/synthetic/surveys.py:139: SettingWithCopyWarning: 
  A value is trying to be set on a copy of a slice from a DataFrame.
  Try using .loc[row_indexer,col_indexer] = value instead
  
  See the caveats in the documentation: http://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
    ) + w

harmonica/tests/test_synthetic_surveys.py::test_ground_survey
harmonica/tests/test_synthetic_surveys.py::test_airborne_survey
harmonica/tests/test_synthetic_surveys.py::test_cut_region_ground_survey
harmonica/tests/test_synthetic_surveys.py::test_cut_region_airborne_survey
  /home/santi/git/harmonica/harmonica/synthetic/surveys.py:142: SettingWithCopyWarning: 
  A value is trying to be set on a copy of a slice from a DataFrame.
  Try using .loc[row_indexer,col_indexer] = value instead
  
  See the caveats in the documentation: http://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
    ) + s
```

Do you have any idea about what's going on here?

harmonica/synthetic/surveys.py
```diff
@@ -0,0 +1,145 @@
...
+Create synthetic surveys for gravity and magnetic observations
+"""
+import pyproj
+from verde import get_region, inside
+from verde.coordinates import check_region
+
+from ..datasets import fetch_britain_magnetic, fetch_south_africa_gravity
+
+
+def airborne_survey(region, cut_region=(-5.0, -4.0, 56.0, 56.5)):
```
leouieda: We might want to set default values for the `region`. For example, keeping it the same as the cut region by default.

leouieda:
```suggestion
def airborne_survey(region, subsection=(-5.0, -4.0, 56.0, 56.5)):
```
Maybe? `cut_region` sounds a bit strange to me.

harmonica/synthetic/surveys.py
```diff
@@ -0,0 +1,145 @@
...
+    # passed region
+    survey = _adecuate_survey(survey, region, cut_region)
+    return survey
+
+
+def ground_survey(region, cut_region=(13.60, 20.30, -24.20, -17.5)):
+    """
+    Create a synthetic ground survey
+
+    The observation points are sampled from the South Africa gravity dataset.
```
leouieda:
```suggestion
    The observation points are sampled from the South Africa gravity dataset
    (see :func:`harmonica.datasets.fetch_south_africa_gravity`).
```

doc/api/index.rst
```diff
@@ -70,6 +70,14 @@ Input and Output
...
     load_icgem_gdf
 
+Synthetic models and surveys
+----------------------------
+.. autosummary::
+   :toctree: generated/
+
+    synthetic.airborne_survey
+    synthetic.ground_survey
+
```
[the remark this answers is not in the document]
leouieda: To be clear, I'm only talking about the submodule. The API section could be the same.

harmonica/synthetic/surveys.py
santisoler: You're right about keeping only the coordinates columns. In fact I ditch all other columns on the last line of `_adecuate_survey` (or `_cut_and_scale` as you propose).

doc/api/index.rst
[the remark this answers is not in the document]
santisoler: I created the new `synthetic` submodule with the intention to store both synthetic surveys and synthetic models.
Because these are not independent datasets I held myself of adding these two functions under `datasets`. 
What do you think?

harmonica/synthetic/surveys.py
```diff
@@ -0,0 +1,145 @@
...
+    datasets.fetch_britain_magnetic:
+        Fetch total-field magnetic anomaly data of Great Britain.
+    """
+    # Sanity checks for region and cut_region
+    check_region(region[:4])
+    check_region(cut_region)
+    # Fetch airborne magnetic survey from Great Britain
+    survey = fetch_britain_magnetic()
+    # Rename the "altitude_m" column to "elevation"
+    survey["elevation"] = survey["altitude_m"]
```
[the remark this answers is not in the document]
santisoler: I thought I've already changed to `survey.filter(...)`. :man_facepalming:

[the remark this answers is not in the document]
santisoler: I thought I've already changed it to `survey.rename(...)` :man_facepalming:

harmonica/synthetic/surveys.py
```diff
@@ -0,0 +1,145 @@
...
+    # Fetch ground gravity survey from South Africa
+    survey = fetch_south_africa_gravity()
+    # Cut the region into the cut_region, project it with a mercator projection to
+    # convert the coordinates into Cartesian and move this Cartesian region into the
+    # passed region
+    survey = _adecuate_survey(survey, region, cut_region)
+    return survey
+
+
+def _adecuate_survey(survey, region, cut_region):
```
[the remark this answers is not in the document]
santisoler: Way better!

harmonica/synthetic/surveys.py
```diff
@@ -0,0 +1,145 @@
...
+    The observation points are sampled from the Great Britain total-field magnetic
+    anomaly dataset. Only a portion of the original survey is sampled and its region is
+    rescaled to the passed ``region``.
+
+    Parameters
+    ----------
+    region : tuple or list
+        Boundaries of the synthetic region where the observation points will be located
+        in the following order: (``east``, ``west``, ``south``, ``north``, ...). All
+        subsequent boundaries will be ignored. All boundaries should be in Cartesian
```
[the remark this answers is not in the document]
santisoler: I agree. The projection to Cartesian coordinates has been inherited from my current EQL research. But keeping in mind that these synthetic datasets may serve as examples for new users, we should respect the entire workflow. And precisely, we expect data to be referenced in geodetic coordinates and not on projected ones.

harmonica/synthetic/surveys.py
santisoler: I hate when I name variables while I'm drafting the new functions and after a few commits I cannot thing a better name for them because I've already got used to the ugly one. `subsection` is way better.

santisoler: What about `subregion`? Or `dataset_subregion`?

harmonica/synthetic/surveys.py
santisoler: I moved the ditching of all the extra columns to each public function, before calling the `_cut_and_scale` function.

santisoler:
@leouieda I think I've solved all the issues.
Instead of setting the region equal to the subsection if it's None, I decided it would be better to don't scale the survey. What do you think about this?

Left some answers on the inline comments.

harmonica/synthetic/surveys.py
leouieda: `dataset_region` to be more explicit? 

> There are 2 hard problems in computer science: cache invalidation, naming things, and off-by-1 errors. -- [Leon Bambrick](https://twitter.com/secretGeek/status/7269997868)

harmonica/synthetic/surveys.py
```diff
@@ -0,0 +1,157 @@
...
+from ..datasets import fetch_britain_magnetic, fetch_south_africa_gravity
+
+
+def airborne_survey(region=None, subsection=(-5.0, -4.0, 56.0, 56.5)):
+    """
+    Create a synthetic ground survey
+
+    The observation points are sampled from the Great Britain total-field magnetic
+    anomaly dataset (see :func:`harmonica.datasets.fetch_britain_magnetic`).
+    Only a portion of the original survey is sampled and its region may be
```
leouieda:
```suggestion
    A portion of the original survey is cut (*data_region*) and the coordinates may be
```

harmonica/synthetic/surveys.py
```diff
@@ -0,0 +1,157 @@
...
+    scaled to the passed ``region``.
```
leouieda:
```suggestion
    scaled to the given *region*.
```

harmonica/synthetic/surveys.py
```diff
@@ -0,0 +1,157 @@
...
+
+    Parameters
+    ----------
+    region : tuple or list (optional)
+        Region at which the survey points coordinates will be scaled.
```
leouieda:
```suggestion
        Survey horizontal coordinates will be scaled to span this area.
```

harmonica/synthetic/surveys.py
```diff
@@ -0,0 +1,157 @@
...
+        The boundaries must be passed in the following order:
+        (``east``, ``west``, ``south``, ``north``, ...), defined on a geodetic
+        coordinate system and in degrees.
+        All subsequent boundaries will be ignored.
```
leouieda:
```suggestion
        Only the 4 horizontal boundaries are used. Subsequent boundaries will be ignored.
```

harmonica/synthetic/surveys.py
```diff
@@ -0,0 +1,157 @@
...
+        If ``None``, the survey points won't be scaled. Default ``None``.
+    subsection : tuple or list (optional)
+        Region where the original Great Britain magnetic dataset will be sampled.
```
leouieda:
```suggestion
        Subsection of the original Great Britain magnetic dataset that will be sampled.
```

harmonica/synthetic/surveys.py
```diff
@@ -0,0 +1,157 @@
...
+        The boundaries must be passed in the following order:
+        (``east``, ``west``, ``south``, ``north``, ...), defined on a geodetic
+        coordinate system and in degrees. All subsequent boundaries will be ignored.
+
+    Returns
+    -------
+    survey : :class:`pandas.DataFrame`
+        Dataframe containing the coordinates of the observation points on a geodetic
+        coordinate system. Longitudes and latitudes are in degrees, and heights in
+        meters.
```
leouieda: Maybe specify the column names?

harmonica/synthetic/surveys.py
```diff
@@ -0,0 +1,157 @@
...
+        ["longitude", "latitude", "height"]
+    )
+    # Cut the survey into the subsection and scale it to the passed region
+    survey = _cut_and_scale(survey, region, subsection)
+    return survey
+
+
+def _cut_and_scale(survey, region, subsection):
+    """
+    Cut the original survey to the subsection and scale it to the passed region
```
leouieda:
```suggestion
    Cut a subsection from the original survey and scale it to the given region.
```

harmonica/synthetic/surveys.py
```diff
@@ -0,0 +1,157 @@
...
+
+    Parameters
+    ----------
+    survey : :class:`pandas.DataFrame`
+        Original survey as a :class:`pandas.DataFrame` containing the following columns:
+        ``longitude``, ``latitude`` and ``height``.
+    region : tuple or list (optional)
+        Region at which the survey points coordinates will be scaled.
```
leouieda:
```suggestion
        Region to which the survey points coordinates will be scaled.
```

harmonica/synthetic/surveys.py
leouieda: :+1:

harmonica/synthetic/surveys.py
santisoler: Yes, it sounds a little better than subsection.

doc/api/index.rst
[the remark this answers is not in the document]
santisoler: I would prefer keeping `datasets` only for real world *data*, such as the ones stored on Rockhound.
And store anything related on synthetic modelling on the new `synthetic` submodule.
We will see how this works in the future, and until `1.0.0` we can change it if we don't like it.

harmonica/synthetic/surveys.py
santisoler: :+1:

doc/api/index.rst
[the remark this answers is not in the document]
leouieda: I bow to your wisdom :bowing_man:

harmonica/synthetic/surveys.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+def airborne_survey(region=None, data_region=(-5.0, -4.0, 56.0, 56.5)):
+    """
+    Create measurement locations for a synthetic airborne survey.
+
+    The observation points are sampled from the Great Britain total-field magnetic
+    anomaly dataset (see :func:`harmonica.datasets.fetch_britain_magnetic`).
+    A portion of the original survey is cut (*data_region*) and the coordinates may be
+    scaled to the given *region*.
```
leouieda: One thing to remember now is to crop these lines at 79 characters.

harmonica/synthetic/surveys.py
```diff
@@ -0,0 +1,159 @@
...
+    passed ``region``.
+
+    Parameters
+    ----------
+    region : tuple or list (optional)
+        Region at which the survey points coordinates will be scaled.
+        The boundaries must be passed in the following order:
+        (``east``, ``west``, ``south``, ``north``, ...), defined on a geodetic
+        coordinate system and in degrees.
+        All subsequent boundaries will be ignored.
```
leouieda: Same change as the above docstring.

harmonica/synthetic/surveys.py
santisoler: You're right. I added it to my TODO list, but forgot to do so.

santisoler: By the way, is there a quick way to do this in vim? If it is, we should include it on the Contributing Guidelines. It's not an automatic way to do so, but at least it's better than nothing.
For example, I managed to do it by entering into VISUAL mode, select the lines I wanted to shrink and used `gq`. Nevertheless, I needed to select each block of the docstring separately instead of selecting the entire docstring and doing it only once.

leouieda: Good question. I did it the same way you did. You could write a macro but it might be more work than doing it manually.

harmonica/synthetic/surveys.py
```diff
@@ -0,0 +1,161 @@
...
+        ["longitude", "latitude", "height"]
+    )
+    # Cut the survey into the data_region and scale it to the passed region
+    survey = _cut_and_scale(survey, region, data_region)
+    return survey
+
+
+def ground_survey(region=None, data_region=(13.60, 20.30, -24.20, -17.5)):
+    """
+    Create a synthetic ground survey
```
leouieda:
```suggestion
    Create measurement locations for a synthetic ground survey
```

harmonica/synthetic/surveys.py
santisoler: I think we will have to do it manually then. Another option is to propose this kind of automation on Black.

leouieda: The problem then would be to have black use different widths for docstrings and code. Which I doubt they would allow :(

leouieda:
This looks good to me :+1: Feel free to merge away when ready :rocket:

PR #127, opened 2019-11-08
Wrap docstrings to 79 characters and configure flake8

### LINKED ISSUE #126: Wrap all docstrings to 79 characters

After fatiando/verde#177 raised by @prisae, fatiando/contributing#9 and fatiando/contributing#10 it has been decided that all docstrings must be wrapped to 79 characters per line.
We should warp all the existing docstrings in Harmonica to 79 characters and also configure flake8 to check for this when running the style checks.

Sadly, there's no way (yet) to automatically change all docstrings, but at least we can use flake8 to raise lines that fail. This can be done by setting `max-doc-length` to 79 characters on the flake8 configuration under `setup.cfg`.

santisoler:
Wrap all docstrings to 79 characters of API functions. Excluded examples
and test functions. Set max-doc-length option to 79 characters in
order to make flake8 able to properly check if any docstring exceeds it.
Wrap comment lines to 79 characters on API functions in order to prevent
flake8 raising an error. Configure flake8 to ignore a docsting larger
than 79 characters on example scripts and test functions. Made minor
modifications on the main line of some docstrings because they couldn't
be splitted into two lines. Made minor modifications on some examples
included on the docstrings of the API functions in order to wrap to 79
characters.

Fixes #126

**Reminders**


santisoler:
I've found that when setting `max-doc-length = 79` on flake8 configuration, it also checks if comments lines are not longer than 79 characters. Can't find a way to configure only docstrings, maybe you @leouieda know how. Although is not too problematic, it's would be strange to write code on 88 characters and comment lines on 79 characters per line.

leouieda:
> Although is not too problematic, it would be strange to write code on 88 characters and comment lines on 79 characters per line.

I'll take it! I didn't find a way of separating comments from docstrings. We can keep this for now. Comments aren't usually huge anyway.

PR #129, opened 2019-11-14
Use napoleon instead of numpydoc to unpin Sphinx

### LINKED ISSUE #128: Replace numpydoc for napoleon to unpin Sphinx

The combination of latest sphinx (2.2.0), numpydoc, and sphinx_rtd_theme
causes a mangled rendering of the parameter list. We were pinning sphinx
to 1.8.5, which is getting old and causing dependency problems.
See discussion in fatiando/pooch#113.

A possible solution is to replace numpydoc with sphinx.ext.napoleon.
It solves the problem although it generates a slightly different output.
The class autosummary template must be modified in order to account for
missing methods summary block.

This replacement has been achieved by @leouieda on fatiando/pooch#122.
See the changes applied on that PR for inspiration.

santisoler:
The combination of latest sphinx (2.2.0), numpydoc, and sphinx_rtd_theme
causes a mangled rendering of the parameter list. We were pinning sphinx
to 1.8.5, which is getting old and causing dependency problems.
Replacing numpydoc with sphinx.ext.napoleon fixes the problem (though
generates slightly different output). Modify autosummary class in order
to generate a Methods Summary. Update Sphinx configuration: change
values of napoleon configuration to compile Returns and Attributes with
the same style as the Parameters list. The changes applied were based on
fatiando/pooch#122.

Fixes #128

**Reminders**


santisoler:
@leouieda I've inspired on fatiando/pooch#122 to make this PR. Although I left a slightly different customization of the autosummary class: on Harmonica we show examples. Besides, I also overwrote  the `numpy_use_ivar` value to True in order to make Attributes lists to compile with the same style as Parameters.

Please check if you would change something else, for example the position of the Examples. I put them at the bottom because if I put them in between the compilation omits a line break and the examples appear at the left of the first method. Like this:

imagen

Finally, wouldn't be nice to standarize the autosummary class template and the sphinx configuration among all libraries?

santisoler:
Nevermind the bad location of the examples, I managed to solve it (and understand it :tada:).

So now I can put the examples on the bottom or before/after the methods summary.
What location looks better?

**Option 1**
- Class docs
- Examples
- Methods Summary
- Methods docs

**Option 2**
- Class docs
- Methods Summary
- Examples
- Methods docs

**Option 3**
- Class docs
- Methods Summary
- Methods docs
- Examples

Issue #68, opened 2019-07-16
Page about the coordinate systems


leouieda:
We need a documentation page describing our coordinate system conventions.

prisae:
I did this recently for empymod. The example (particularly the coordinate-system figure) might be useful to get this issue going:

https://empymod.readthedocs.io/en/stable/examples/coordinate_system.html

santisoler:
Thanks @prisae for the motivation. We should have tackled this issue long time ago, so this could be a good opportunity to do so.

PR #136, opened 2019-12-26
Interpolate data using EQL in spherical coordinates

### LINKED ISSUE #94: Enhance EQL gridder to interpolate data in spherical coordinates

Would be nice to improve the EQL Harmonic gridder to be able to interpolate data in spherical coordinates (without the need to project it). This feature would be very useful when gridding potential data on regional or continental scales where the curvature of the Earth must be taken into account. It can be done by modifying the `EQLHarmonic` class, allowing to pass a `coordinate_system` argument which will regulate how to compute the Green's function (1 / distance) between the point sources and the observation points.

santisoler:
Add new `EQLHarmonicSpherical` class to interpolate data in spherical
coordinates using the equivalent layer technique. Intended to be used on large
scale data, where the curvature of the Earth must be taken into account in
order to produce accurate interpolations.

Fixes #94

**Reminders**


santisoler:
@leouieda I drafted a possible solution to #94. Nevertheless, I'm skeptical about using the `coordinate_system` argument for specifying whether we are passing coordinates in Cartesian or spherical. Wouldn't be better to have separated classes for each coordinate system? Maybe this will create some repeated code, but would ease their usage (passing the `coordinate_system` argument could be easily forgettable).

leouieda:
>  Wouldn't be better to have separated classes for each coordinate system?

I think so. You could use the same code you have here but have the `EQLHarmonicSpherical` (or something like that) set the `greens_function` attribute.

santisoler:
Agree. I'm thinking we could create a `BaseEQLHarmonic` class not intended for use it, and then two subclasses: `EQLHarmonicCartesian` and `EQLHarmonicSpherical`. What do you think? It worth the effort or it would be better to just append the `EQLHarmonicSpherical` as child of the current `EQLHarmonic` with just a different `green_function` attribute?

santisoler:
I created a new `EQLHarmonicSpherical` class as a child of `EQLHarmonic`. The main difference between them is the attribute `greens_function` which points to different Numba jitted functions for computing the Green's function on each case.

@leouieda I added the `greens_function` as an instance attribute because I'm having some problems setting it as a class attribute. It seems that when it's called (inside `jacobian`, for example), the instance is passing 7 arguments instead of 6 (maybe it's passing the instance itself as first argument?).
If you have a solution to this let me know. Otherwise I would keep it as an instance attribute.

harmonica/equivalent_layer/harmonic.py
```diff
@@ -186,23 +197,197 @@ def jacobian(
...
+    Parameters
+    ----------
+    damping : None or float
+        The positive damping regularization parameter. Controls how much
+        smoothness is imposed on the estimated coefficients.
+        If None, no regularization is used.
+    points : None or list of arrays (optional)
+        List containing the coordinates of the point sources used as the
+        equivalent layer. Coordinates are assumed to be in the following order:
+        (``longitude``, ``latitude``, ``radius``). Both ``longitude`` and
```
leouieda: Should we differentiate latitude from *spherical latitude* here?

leouieda: Maybe not since we say it's spherical in the docstring and we didn't do this of the forward modeling functions, right?

leouieda:
>  I added the greens_function as an instance attribute because I'm having some problems setting it as a class attribute. It seems that when it's called (inside jacobian, for example), the instance is passing 7 arguments instead of 6 (maybe it's passing the instance itself as first argument?).
If you have a solution to this let me know. Otherwise I would keep it as an instance attribute.

@santisoler I don't know what could have gone wrong but it's fine the way it is. This is not critical.

Overall this looks great! I'd be fine with merging this in and working on an example in a followup PR but it's fine if you want to add it here still.

harmonica/equivalent_layer/harmonic.py
santisoler: I agree with that. If we decide to change it to spherical latitude here, we must do it elsewhere. But I don't think this is critical. The `radius` coordinate next to it would make it sufficiently clear to any usual user of Harmonica that it's asking spherical coordinates. And for any newcomer I think we should make this very clear on the _Page about coordinate systems_ (#68).

santisoler:
I would like to add an example for this. But maybe it's better to leave it for another PR where we can discuss it.
I'm thinking to grid the entire South Africa observed gravity dataset, but because it would take a lot of memory and time I'm considering to randomly choose a not so large number of points. It's not the way we should do it on a real world scenario, but it could serve for this example. Another option would be to find a different survey with less points but with regional scale. What do you think?

harmonica/equivalent_layer/harmonic.py
leouieda: Right, that's probably better

leouieda:
> I'm thinking to grid the entire South Africa observed gravity dataset, but because it would take a lot of memory and time I'm considering to randomly choose a not so large number of points. It's not the way we should do it on a real world scenario, but it could serve for this example. Another option would be to find a different survey with less points but with regional scale. What do you think?

If you want to downsample, then use `verde.BlockMean` with the final grid resolution (which you can make a bit coarse for this example).

PR #152, opened 2020-03-04
Add example for EQLHarmonicSpherical

santisoler:
Add example for `EQLHarmonicSpherical` using South Africa gravity dataset.
Original data is downsampled to speed-up computations and then converted to
spherical coordinates to fit the gridder. Then the data is predicted on
a regular grid defined on geodetic coordinates by passing
`Ellipsoid.geodetic_to_spherical` as the `projection` argument.

Related to #94 and #136

**Reminders**


examples/eql/harmonic_spherical.py
```diff
@@ -0,0 +1,121 @@
+"""
+Gridding in Spherical coordinates
```
leouieda:
```suggestion
Gridding in spherical coordinates
```

examples/eql/harmonic_spherical.py
```diff
@@ -0,0 +1,121 @@
...
+=================================
+
+When gridding magnetic or gravity data on large regions thought the equivalent
+layer technique, the curvature of the Earth must be taken into account.
+Projecting data for large regions may introduce projection errors on the
+predictions.
```
leouieda:
```suggestion
The curvature of the Earth must be taken into account when gridding and 
processing magnetic or gravity data on large regions. In these cases, projecting 
the data may introduce errors due to the distortions caused by the projection.
```

examples/eql/harmonic_spherical.py
```diff
@@ -0,0 +1,121 @@
...
+
+:class:`harmonica.EQLHarmonicSpherical` is able to estimate the
+coefficients of a set of point sources (the equivalent layer) that fit the
+observed harmonic data given in Spherical coordinates.
+It allows to benefit from the same advantages that the equivalent layer
+technique offers while taking into account the curvature of the Earth.
```
leouieda:
```suggestion
:class:`harmonica.EQLHarmonicSpherical` implements the equivalent layer technique
in spherical coordinates. It has the same advantages as the Cartesian equivalent layer
(:class:`harmonica.EQLHarmonic`) while taking into account the curvature of the Earth.
```

examples/eql/harmonic_spherical.py
```diff
@@ -0,0 +1,121 @@
...
+"""
+import numpy as np
+import boule as bl
+import verde as vd
+import harmonica as hm
+import cartopy.crs as ccrs
+import matplotlib.pyplot as plt
```
leouieda:
```suggestion
import numpy as np
import cartopy.crs as ccrs
import matplotlib.pyplot as plt
import boule as bl
import verde as vd
import harmonica as hm
```

I like to put our packages last to call more attention to them :slightly_smiling_face:

examples/eql/harmonic_spherical.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+# Fetch the sample gravity data from South Africa
+data = hm.datasets.fetch_south_africa_gravity()
+
+# Downsample the number of points to speed-up the computations for this example
+reducer = vd.BlockMean(spacing=0.2, drop_coords=False)
+(longitude, latitude, elevation), gravity_data, _ = reducer.filter(
+    (data.longitude, data.latitude, data.elevation), data.gravity,
+)
```
leouieda:
```suggestion
# Downsample the data using a blocked mean to speed-up the computations 
# for this example. This is preferred over simply discarding points to avoid 
# aliasing effects.
blocked_mean = vd.BlockReduce(np.mean, spacing=0.2, drop_coords=False)
(longitude, latitude, elevation), gravity_data = blocked_mean.filter(
    (data.longitude, data.latitude, data.elevation), data.gravity,
)
```

I usually go for `BlockReduce` in these cases to avoid having it return the weights (and the ugly `_` hack).

examples/eql/harmonic_spherical.py
```diff
@@ -0,0 +1,121 @@
...
+
+# Compute gravity disturbance by removing the gravity of normal Earth
+ellipsoid = bl.WGS84
+gamma = ellipsoid.normal_gravity(latitude, height=elevation)
+gravity_disturbance = gravity_data - gamma
+
+# Convert data coordinates to spherical
```
leouieda:
```suggestion
# Convert data coordinates from geodetic (longitude, latitude, height) to 
# spherical (longitude, spherical_latitude, radius).
```

examples/eql/harmonic_spherical.py
```diff
@@ -0,0 +1,121 @@
...
+
+# Interpolate data on a regular grid with 0.2 degrees spacing defined on
+# geodetic coordinates. To do so we need to specify we want grid coordinates to
+# be converted to spherical geocentric coordinates before the prediction is
+# carried out. This can be done though the projection argument. The
+# interpolation requires an extra coordinate (upward height). By passing in
+# 2500 m above the maximum observation radius, we're effectively
+# upward-continuing the data (maximum height of observation points is 2400 m).
+# All the parameters passed to build the grid (region, spacing and
+# extra_coords) are in geodetic coordinates.
```
leouieda:
```suggestion
# Interpolate data on a regular grid with 0.2 degrees spacing defined on
# geodetic coordinates. To do so we need to specify that we want coordinates to
# be converted to spherical geocentric coordinates before the prediction is
# carried out. This can be done though the "projection" argument. 
# The interpolation requires an extra coordinate (upward height). By passing in
# 2500 m above the ellipsoid, we're effectively
# upward-continuing the data (maximum height of observation points is 2400 m).
# All the parameters passed to build the grid (region, spacing and
# extra_coords) are in geodetic coordinates.
```

examples/eql/harmonic_spherical.py
santisoler: Ok, agree. I tend to list packages in increasing line length order, it just please my eyes haha.

examples/eql/harmonic_spherical.py
santisoler: Oops, I forgot to change the reference of the elevation.

santisoler:
Thanks for the review @leouieda 

I think this is good to be merged, although maybe we would like to wait until fatiando/verde#237 is merged (and a new release of Verde is out) so we can replace the `vd.distance_mask` for the new `vd.convexhull_mask`. Leave it up to you.

leouieda:
No need, we can always update it after the release. This works fine the way it is now.

PR #183, opened 2020-07-29
WIP Add option to change jacobian dtype on EQL gridders

### LINKED ISSUE #176: Allow changing data type of EQLs Jacobian elements

We know that fitting an EQL gridder is a process that requires a fair amount of memory, mainly to store the Jacobian matrix.
One small step into reducing the required memory would be to change the data type used to store the Jacobian elements, which are defined as `float64` by default.
Currently, the EQLs offer no option to change that, even though the `jacobian` method has an optional argument `dtype` to change it.
Would be nice to have a `jac_dtype` argument that could be passed to the EQL constructor in order to change the representation of the Jacobian elements.
Although the memory consumption won't drop by a lot, users may reduce it by half storing the elements as `float32`.
I suspect that using `float32` won't significantly compromise the accuracy of the predictions.

santisoler:
Add an extra parameter on `EQLHarmonic` and `EQLSpherical` constructors to change the dtype of Jacobian elements.
This may help on reducing memory consumption while might drop accuracy of the predicted coefficients.

Fixes #176

**Reminders**:

### DIFF
--- a/harmonica/equivalent_layer/harmonic.py
+++ b/harmonica/equivalent_layer/harmonic.py
@@ -71,6 +71,10 @@ class EQLHarmonic(vdb.BaseGridder):
         this constant *relative_depth*. Use positive numbers (negative numbers
         would mean point sources are above the data points). Ignored if
         *points* is specified.
+    jacobian_dtype : str, type or numpy dtype
+        Variable type used for defining Jacobian elements. Use
+        ``"float32"`` for reducing memory consumption, but you might expect
+        some accuracy on fitted coefficients. Default ``"float64"``.
 
     Attributes
     ----------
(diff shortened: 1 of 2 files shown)


santisoler:
I'm trying to test the accuracy of the EQL gridder that uses `jacobian_dtype="float32"` by:
1. Checking if it can recover the data on the observation points,
2. Comparing predicted values on a grid with an EQL that uses `jacobian_dtype="float64"`.

And I'm facing some drop on the accuracy (an `rtol` of ~0.5), which by visual inspection is not huge, but can make `npt.assert_allclose` to fail for some survey points distributions.

So my main question is: should we add this feature?

If yes, which is the best way to test it?

A few ideas came to my head, but I would like to hear some opinions:
1. The accuracy problem is probably originated while comparing tiny values (close to zero).
2. We can then shift the data, so there are no values close to zero.
3. We can compare them through absolute difference (`atol`), rather than with relative difference (`rtol`).

I've tested some of these ideas: 2 and 3 seem to work (but I'm not sure if they are a proper way to test this feature), although they might get problems for a certain random choice of the coordinate points. So another option is to fix the choice of those coordinate points by specifying the `random_state` on `vd.scatter_points`.

leouieda:
This needs a lot of work on Verde to allow user input of dtypes throughout the project. There are several places where arrays are created that need to take dtype as input.

santisoler:
I don't think that changing the dtype of the Jacobian would produce any significant improvement on the memory saving side. In fact, as I saw, the accuracy of the predictions can drop below the desired level. I'm closing this PR then.

Issue #179, opened 2020-07-17
Remove sample datasets after moved to Rockhound


santisoler:
**Description of the desired feature**

On the last community call we decided to move all sample datasets in Fatiando to Rockhound.
Harmonica datasets are being moved to Rockhound on https://github.com/fatiando/rockhound/pull/84.
After it is merged, we should remove the harmonica.datasets along with the data folder. Besides, all gallery examples should download their data from Rockhound, therefore it should be added as an optional dependency.

**Are you willing to help implement and maintain this feature?** Yes, but would leave to anyone that wants to tackle it.

lperozzi:
This means that 

```python notebook
# Read South Africa topography
south_africa_topo = hm.datasets.fetch_south_africa_topography()
```
will not work anymore right? I tried to fetch it in a environment with the last `prisms-layer` branch installed, but I have the following error:

```python notebook
HTTPError: 404 Client Error: Not Found for url: https://github.com/fatiando/harmonica/raw/master/data/south-africa-topography.nc.xz

```

santisoler:
Sorry for the delay! The notification was buried among others.

Yes, we probably end up ditching the `harmonica.datasets` module, but I don't expect to do that before the next release.

The issue with the South Africa topography raises because that dataset hasn't been merged to `master` yet. That's why you get that error. A quick solution in order to run the tests from data branch is to manually copy the `south-africa-topography.nc.xz` file to the default cache folder of Harmonica. This way, when the dataset is fetched through `hm.datasets.fetch_south_africa_topography()`, Harmonica won't try to download it because it's already on the cache folder.

Issue #124, opened 2019-11-08
Repeated points on sample datasets downloaded from ICGEM


santisoler:
**Description of the problem**

We currently have three sample datasets that have been downloaded from the [ICGEM Calculation Service](http://icgem.gfz-potsdam.de/calcgrid):
- Earth Topography
- Earth Gravity
- Earth Geoid

Each of these models has been created on a regular grid in a geodetic coordinate system.
On the latitude direction they span between -90 and 90 degrees, while on the longitudinal one they do so between -180 and 180 degrees.
The problem is that the -180 and 180 longitudinal degrees are equivalent, therefore each dataset contains one row of repeated values.

Would be nice to regenerate and upload these datasets without any repeated values.
What has to be done?
1. Fetch each dataset with the existing functions (e.g. `hm.datasets.fetch_topography_earth()`).
2. Remove the repeated longitudinal coordinate and its corresponding data values.
3. Save the new dataset to a NetCDF file (`.nc`)
4. Compress it to a `.xz` file.
5. Replace the existing files on `data` with the new ones.
6. Get the sha256 hash of the new file and update `harmonica/datasets/registry.txt`

leouieda:
Alternatively, we could keep the datasets as they are and only drop the point when we load them.

santisoler:
Since we are moving every dataset to Ensaio, I'm closing this issue.

PR #184, opened 2020-08-03
Keep metadata of sample datasets in the xarray objects

### LINKED ISSUE #125: Missing attributes on ICGEM datasets
**Description of the problem**

We currently have three sample datasets that have been downloaded from the [ICGEM Calculation Service](http://icgem.gfz-potsdam.de/calcgrid):
- Earth Topography
- Earth Gravity
- Earth Geoid

Each dataset can be fetched through its own function. Every one of these functions claims to have more details regarding the generation of the grid under the `attrs` attribute of the `Dataset`.
The problem is neither of the returned `Dataset`s have non empty `attrs`.

Would be nice to correct this and include the attributes related to the grid generation.
I think these grids where generated by @leouieda, so maybe he has more idea of what information should be included on the `attrs`.

Just by naming some possible attributes that appear on the ICGEM Calculation Service:
- Reference ellipsoid
- Height above the ellipsoid
- Grid step
- Model used (for gravity and geoid)
- Tide system
- Zero degree term (yes/no)
- Gaussian filter and filter length
- Low-pass filtering
    - Start gentle cut
    - Maximum degree

nshea3:
Fixes #125 

Many global attributes are included in the netcdf files, however they are dropped in the `fetch_*` functions in `sample_data.py` with the float64 cast: `data = xr.open_dataset(fname, engine="scipy").astype("float64")`.  This is a known issue in xarray (https://github.com/pydata/xarray/issues/2049) and the current solution is to open the NetCDF file as-is, cast to the new datatype, and then copy the global attributes to the new object.

Also added tests to `test_sample_data.py` to check that the `attrs` attribute is not an empty dictionary. 

**Reminders**:

- [x] Add new public functions/methods/classes to `doc/api/index.rst` and the base `__init__.py` file for the package.
- [x] Write detailed docstrings for all functions/classes/methods. It often helps to design better code if you write the docstrings first.
- [x] Add your full name, affiliation, and [ORCID](https://orcid.org) (optional) to the `AUTHORS.md` file (if you haven't already) in case you'd like to be listed as an author on the [Zenodo](https://zenodo.org/communities/fatiando) archive of the next release.


santisoler:
Since we are about to move all datasets to Rockhound (see #179), we should remember to apply the same changes we perform here on https://github.com/fatiando/rockhound/pull/84.

nshea3:
Excellent point thank you @santisoler I will make that change so we avoid duplicating the xarray.

I'll add a test for the `refsysname` attribute as well. Are there any others I should include?

santisoler:
> I'll add a test for the refsysname attribute as well. Are there any others I should include?

Maybe we can add those attributes of which the data depends on.
For example, the `refsysname` is crucial because it sets the reference frame (and which normal gravity was removed if we are requesting for gravity anomalies or disturbances).
I would also test the `modelname`, the `max_used_degree` and probably the `tide_system`.
Another things we might want to test if the min and maximum values agree with the ones that are hardcoded in the `tests/sample_data.py`:

https://github.com/fatiando/harmonica/blob/2980071fce37948f2d13d8b7120c96ed18923410/harmonica/tests/test_sample_data.py#L35-L42

Feel free to add as many tests as you think, more tests is usually a good thing to have!

nshea3:
Hi @santisoler, I have made the change you suggested, the `attrs` are now copied to a dictionary and set after the conversion. 

I have also added several tests for the presence of specific attributes: `refsysname`, `max_used_degree`, `modelname`, and `tide_system`.

I have not yet changed the netcdf files as discussed in #124. I started editing the files with the `netcdf4-python` library but I realized that the global attributes do not update as changes are made to the files, so the global attributes would be incorrect.

I am currently trying to regenerate the files from the ICGEM website (http://icgem.gfz-potsdam.de/calcgrid) without the repeated longitudes but the only output option is a `.gdf` file and I do not know how to convert that to a `.cdl` or `.nc` file?

santisoler:
@nshea3 Sorry for stalling this PR for too long. I don't know why it kept under my radar for too long.
I'll merge right away if you agree.

We are planning a new Harmonica release for the next week(s). Thanks for the contribution!

PR #186, opened 2020-09-02
Add function to create a layer of prisms

santisoler:
Add function that creates a layer of prisms as a `xr.Dataset` which contains
the coordinates of the center of each prism, its top and bottom boundaries (as
xarray coordinates) and each property (as density, magnetization, etc) as data
array. Add `PrismsLayer` class as a [dataset accesor](https://xarray.pydata.org/en/stable/internals.html#extending-xarray)
for `xr.Dataset`s, which allows to define several attributes, properties and
methods for the prism layer, which can be accessed through the `prisms_layer`
attribute of the `xr.Dataset`. Add a `gravity` method that computes
gravitational fields generated by every prism in the layer whose top boundary,
bottom boundary or density are not `np.nan`s.
Add two gallery examples for the prisms layer, and a new South Africa
topography dataset, which is a downsample of ETOPO1.

This PR may serve as a prototype for solving #83, but for prisms. The one for tesseroids could be equivalent to this one, but with some minor checks due to longitude continuity.

**Reminders**:


santisoler:
I suspect that ETOPO1 is referenced on the geoid (need to check), so we might want to add the geoid height (from the reference ellipsoid) to the topography + bathymetry grid before creating the terrain layer.

santisoler:
@lperozzi Would you like to review this PR?

lperozzi (commented):
I will add another points that I've remarked today. If the aim is to extract the topographic correction for the observations stations from the prisms gravity forward, we should be careful with the bord effects.

santisoler:
Thanks for the feedback @lperozzi!

> I will add another points that I've remarked today.

Nice! Can't wait to see them!

santisoler:
> If the aim is to extract the topographic correction for the observations stations from the prisms gravity forward, we should be careful with the bord effects.

I totally agree, maybe we should change the gallery example so it takes a larger region than the survey area for building the prisms. Thanks for noticing it!

examples/forward/prisms_topo_gravity.py
```diff
@@ -0,0 +1,76 @@
...
+# Make a plot of the computed gravity
+plt.figure(figsize=(7, 6))
+ax = plt.axes()
+#  ax = plt.axes(projection=ccrs.Mercator())
+tmp = ax.pcolormesh(*coordinates[:2], prisms_gravity)
+ax.set_title("Terrain gravity effect")
+plt.colorbar(tmp, label="mGal")
+ax.set_aspect("equal")
+plt.tight_layout()
+plt.show()
```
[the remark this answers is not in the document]
santisoler: Sure! You'll probably want to apply some gridding or downsampling to your topographic grid in order to build the prisms: there's no point to use a very high resolution DEM to apply terrain correction, while if you have some topographic height measurements and low resolution DEM you can combine them using a gridder. But those tasks should be carried out by the user.

harmonica/forward/prisms_layer.py
```diff
@@ -0,0 +1,331 @@
...
+        data_vars = {
+            prop: (dims, value * np.ones_like(coordinates[0]))
+            for prop, value in properties.items()
+        }
+    attrs = {"coords_units": "meters", "properties_units": "SI"}
+    return xr.Dataset(data_vars=data_vars, coords=coords, attrs=attrs)
+
+
+@xr.register_dataset_accessor("prisms_layer")
+class PrismsLayer:
```
santisoler, line 154: Rename it to something like `XarrayDataAccessorPrismsLayer`

santisoler:
Would be nice to use the new `make_xarray_grid` for creating the prisms layer. We could also wait for https://github.com/fatiando/verde/pull/300 to be merged so we will be able to pass the `easting` and `northing` coordinates either as 1d or 2d arrays.

santisoler:
@lperozzi @leouieda Can you take a look at the latest changes I made to the prisms layer?

Now the prisms layer must be defined by passing the `coordinates` (easting and northing as 1d arrays, so we can use preexisting xarray.Dataset), a `surface` 2d array and a `reference` which can be either a 2d array or a float. The `top` and `bottom` coordinates are created by the `DatasetAccessorPrismsLayer.update_top_bottom` method, which can be used when needed, not only when we are building the layer (useful for inversions). I removed the method to compute the gravity to avoid having multiple ways of achieving the same goal, so users will have to use the regular `hm.prism_gravity` function by passing the prisms boundaries and their density, which can be easily extracted from the layer.

At last, I updated the tests and the examples. Feel free to review it and suggest any change.

PS: I would like to add a terrain correction example, including the whole process (start from gravity observations, compute gravity disturbance, grid the data, apply terrain correction), but we could leave it for another PR.

santisoler:
@leouieda I'm trying to finish this PR. I think it's almost ready, but I'm using the prisms_layer feature for the Transform21 tutorial and I found that it can create some problems when the surface array has nans.

I decided to be conservative and raise and error if any nan is present on `surface` or `reference`.

Another option is to have a `drop_nans=False` optional argument that replaces the nans on the `surface` for the respective value of the `reference`, thus generating a prisms without volume.

What do you think?

santisoler:
> @leouieda I'm trying to finish this PR. I think it's almost ready, but I'm using the prisms_layer feature for the Transform21 tutorial and I found that it can create some problems when the `surface` array has `nan`s.
> 
> I decided to be conservative and raise and error if any nan is present on `surface` or `reference`.
> 
> Another option is to have a `drop_nans=False` optional argument that replaces the nans on the `surface` for the respective value of the `reference`, thus generating a prisms without volume.
> 
> What do you think?

We decided to allow `surface` array with nans and add a `gravity` method that computes the gravity field generated by the layer, ignoring the masked points.

harmonica/forward/prisms_layer.py
```diff
@@ -0,0 +1,411 @@
...
+
+@xr.register_dataset_accessor("prisms_layer")
+class DatasetAccessorPrismsLayer:
+    """
+    Defines dataset accessor for layer of prisms
+
+    .. warning::
+
+        This class is not intended to be initialized.
+        Use the :func:`prisms_layer` accessor for accessing the methods and
```
santisoler, line 141:
```suggestion
        Use the `prisms_layer` accessor for accessing the methods and
```

harmonica/forward/prisms_layer.py
```diff
@@ -0,0 +1,411 @@
...
+
+    @property
+    def dims(self):
+        """
+        Return the dims tuple of the prism layer
+
+        The tuple follows the xarray order: ``"northing"``, ``"easting"``.
+        """
+        return ("northing", "easting")
+
```
santisoler, line 160:
```suggestion
```

harmonica/forward/prisms_layer.py
```diff
@@ -0,0 +1,411 @@
...
+        -------
+        mask : 2d-array
+            Array of bools that can be used as a mask for selecting prisms with
+            no nans on top and bottom boundaries.
+        """
+        nonans_top = np.logical_not(np.isnan(self._obj.top.values))
+        nonans_bottom = np.logical_not(np.isnan(self._obj.bottom.values))
+        return np.logical_and(nonans_top, nonans_bottom)
+
+    def get_prisms(self):
```
leouieda, line 363:
```suggestion
    def to_prisms(self):
```

This fits better with the rest of the xarray methods: http://xarray.pydata.org/en/stable/pandas.html

leouieda: Could keep this private for now and provide it after if desired.

harmonica/forward/prisms_layer.py
```diff
@@ -0,0 +1,411 @@
...
+        See also
+        --------
+        harmonica.prism_gravity
+        """
+        boundaries = self.get_prisms()
+        # Get the mask for selecting only the boundaries of the prisms that
+        # have no nans on their top and bottom
+        nonans_boundaries = self._get_nonans_prisms_mask().ravel()
+        # Get density array
+        density = getattr(self._obj, density_name).values.ravel()
```
leouieda, line 331:
```suggestion
        density = self._obj[density_name].values.ravel()
```
Works for density names that aren't valid Python variable names.

santisoler:
I think this is ready to go. I'll merge it even the CIs are failing: the docs build is failing because it can't download the new South Africa topography dataset. If `master` fails after merge, I'll fix it right away.

PR #190, opened 2020-09-22
Improve EQL harmonic classes

### LINKED ISSUE #163: Add upward argument to EQLHarmonic grid and scatter methods

Would be better to explicitly ask for the upward coordinate when trying to use EQL to predict on grid and scatter points. Currently it must be passed as the keyword argument `extra_coords`, which is not explicitly present in Harmonica documentation.

santisoler:
Make `EQLHarmonic` and `EQLHarmonicSpherical` independent subclasses of Verde's
`BaseGridder`. Add positional upward argument to `grid` and and `profile`
methods of EQLHarmonic. Raise a `NotImplementedError` when calling the
`scatter` method because it will be deprecated on Verde v2.0.0. Raise
a `NotImplementedError` when calling `EQLHarmonicSpherical.profile` method: the
profile should be created out of great-circle distances rather than absolute
differences of longitude and latitude angles. Remove the `projection` argument
on `EQLHarmonicSpherical.grid` method because it's intended to be used only on
Cartesian gridders. Reduce tests on EQLs with Numba disabled to speed them up.
Add test functions to catch `NotImplementedError`s and to check if `projection`
is not a valid argument of `EQLHarmonicSpherical.grid` method.

Fix #163

**Reminders**:


santisoler:
@leouieda Would you take a look at this? As we discussed, I decided to split both `EQLHarmonic` and `EQLHarmonicSpherical` into separated classes, childs of `BaseGridder`. Besides I added the `upward` positional argument to the `grid` and `profile` methods, while the `scatter` and   `EQLHarmonicSpherical.profile` methods raise a `NotImplementedError`.

I've also managed to reduce the tests that run with Numba disabled. Now the `make test_coverage` runs in ~7s instead of ~30s. :rocket:

PR #196, opened 2020-10-13
Replace versioneer with setuptools_scm

santisoler:
Replace `versioneer` with `setuptools_scm` for getting current version of
Harmonica. `setuptools_scm` doesn't require to store additional files to work,
it can be installed and used through `setup.py`. Remove all `versioneer`
related files and mentions. Add `setuptools_scm` to `requirements.txt` and
`environment.yml`. Replace `setup.cfg` for `.flake8`. 
Set `local_scheme = "no-local-version"` on `deploy.yml` GitHub Action before
publishing to Test PyPI.

Related to fatiando/maintenance#4


.github/workflows/deploy.yml
```diff
@@ -62,12 +62,9 @@ jobs:
...
       - name: Build source and wheel distributions
         run: |
-          # Change the versioneer format to "pre" so that the commit hash isn't
-          # included (PyPI doesn't allow it). Can't do this permanently because
-          # we rely the hash to indicate to the tests that this is a local
-          # verison instead of a published version.
-          sed --in-place "s/pep440/pep440-pre/g" setup.cfg
-          python setup.py sdist bdist_wheel
+          # Use the no-local-version configuration on setuptools_scm because
+          # is the only one compatible with PyPI.
```
santisoler, line 66:
```suggestion
          # is the only one compatible with Test PyPI.
```

harmonica/__init__.py
```diff
@@ -11,7 +11,7 @@
 from .forward.prism import prism_gravity
 from .equivalent_layer.harmonic import EQLHarmonic, EQLHarmonicSpherical
 
-# Get the version number through versioneer
+
```
santisoler, line 14:
```suggestion
# Get the version number through setuptools_scm
```

setup.py
```diff
@@ -54,13 +51,30 @@
 ]
 PYTHON_REQUIRES = ">=3.6"
 
+# Configuration for setupttools_scm
```
santisoler, line 54:
```suggestion
# Configuration for setuptools_scm
```

.github/workflows/continuous-integration.yml
```diff
@@ -68,7 +68,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
         with:
-          # Need to fetch more than the last commit so that versioneer can
+          # Need to fetch more than the last commit so that setuptools_scm can
```
santisoler, line 71:
```suggestion
          # Need to fetch more than the last commit so that setuptools-scm can
```

santisoler:
Update installing instructions for latest development version to use `pip install git+https://github.com/fatiando/harmonica` instead of `pip install https://...`.

.github/workflows/deploy.yml
```diff
@@ -62,12 +62,9 @@ jobs:
...
         run: |
-          # Change the versioneer format to "pre" so that the commit hash isn't
-          # included (PyPI doesn't allow it). Can't do this permanently because
-          # we rely the hash to indicate to the tests that this is a local
-          # verison instead of a published version.
-          sed --in-place "s/pep440/pep440-pre/g" setup.cfg
-          python setup.py sdist bdist_wheel
+          # Use the no-local-version configuration on setuptools-scm because
+          # is the only one compatible with Test PyPI.
+          HARMONICA_VERSION_LOCAL_SCHEME="no-local-version" python setup.py sdist bdist_wheel
```
santisoler, line 67: Instead of using an env variable, it's better to apply the same hack Leo was doing, mainly because once we move away from `setup.py` to `pyproject.toml` we won't have the option to check if an env variable exists or not.

santisoler: We are going to set `local_scheme = "no-local-version"` by default, so no shananginan is required.

harmonica/version.py
```diff
@@ -1,9 +1,11 @@
...
-from ._version import get_versions
+from pkg_resources import get_distribution
 
 
-full_version = get_versions()["version"]
-git_revision = get_versions()["full-revisionid"]
+# Get semantic version through setuptools-scm
+version = get_distribution("harmonica").version
+# Append a "v" before the semver version so it looks like: v0.1.0
+full_version = "v" + version
```
santisoler, line 11: Do you agree with this @leouieda?

doc/conf.py
```diff
@@ -80,7 +80,7 @@
 year = datetime.date.today().year
 project = "Harmonica"
 copyright = "2018-{}, The Harmonica Developers".format(year)
-if len(full_version.split("+")) > 1 or full_version == "unknown":
+if "dev" in full_version:
     version = "dev"
 else:
     version = full_version
```
santisoler, line 86: Now that we are using `no-local-version` by default, we can simplify how we detect if we are running a development version or a released one. Do you like it @leouieda ?

santisoler:
> After discussing this with @leouieda, we decided to set `local_scheme = "no-local-version"` by default. This makes easier to upload to Test PyPI without any shananigans on the Actions workflow and we actually don't need the version to have the commit hash or a date, just to know that we are on an untagged commit. More information can be obtained with `git status`.
> 
> After this decision, `setuptools_scm` will raise error if we try to install the dev version of Harmonica though pip in the following way:
> `pip install https://github.com/fatiando/harmonica`.
> 
> Instead, we would need to do so with:
> 
> `pip install git+https://github.com/fatiando/harmonica`
> 
> But this requires `git` to be installed (it clones the repository instead of downloading a tarball).
> 
> So we need to update installing instructions for latest development version to use `pip install git+https://github.com/fatiando/harmonica` instead of `pip install https://...`, and add a note that this step now requires `git` installed.

So it seems that not adding the local version string gets in conflict with Pooch, specifically on detecting if Pooch should use `version` or `version_dev` for creating the cache folder:

https://github.com/fatiando/pooch/blob/1f60c6ffe143a233bf15b5d1d1ac5acc648504ba/pooch/core.py#L380-L381

https://github.com/fatiando/pooch/blob/1f60c6ffe143a233bf15b5d1d1ac5acc648504ba/pooch/utils.py#L108-L151

I'll try to restore the default behaviour of `local_scheme` and add some shananingans with `sed` on the Actions workflow for publishing to TestPyPI.

doc/install.rst
```diff
@@ -68,6 +68,7 @@ there::
     pip install .
 
 
+
```
santisoler, line 71:
```suggestion
```

harmonica/version.py
santisoler: I'll add the `"v"` before the version number in order to keep the same style as we are using now in Fatiando.
If we decide to change it we can do it afterwards.

PR #198, opened 2020-10-26
WIP Formulating and testing Pratt's isostatic calculation 

andieie:
Greetings,

This is a test pull request and the first time opening a pull request in this project. 

I created isostasy_pratt.py closely following the existing isostasy.py format that @aguspesce did. In the new file I included the following :
- documentation of the pratt calculation and hypothesis. 
- function that calculates density of roots by inputing: topography, density of water, density of crust, and compensation depth.


andieie:
Hola @santisoler y @aguspesce this is the error I get in line 79: `IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices`

As I have mentioned before, I am learning how xarray allocates things. Hopefully what I am doing can be written another way or fixed with other variables.

santisoler:
Hi @andieie! Congratulations for your first PR! :tada: :confetti_ball:

I really like the docstring, it's very explicit and very nice to follow by anyone not so experienced with Pratt isostasy.

There are some pieces of the code that need some work though, let me help you with that.

First, there's no need for `isostasy_pratt()` to live inside its own file, I think we should move the function to `harmonica/isostasy.py`, where `isostasy_airy()` lives.

Secondly, I saw you added an example to test how the implementation works. That's a great way to start designing! :clap: 
I would recommend you to create a new `examples/isostasy_pratt.py` file just for the example. And, instead of copying the code inside the function, make the example to actually use it, as any user would use the new function once it's merged.
One minor detail: for the example to be able to use the function like:
```python
import harmonica as hm

hm.isostasy_pratt(...)
```
you need to import the function on `harmonica/__init__.py`. See how the `isostasy_airy` is imported:

https://github.com/fatiando/harmonica/blob/371fff90af5fc4f609ee74c7ef6621f5cc4370d5/harmonica/__init__.py#L7

Regarding the error message you get, I don't exactly know how are you getting it, because the `isostasy_pratt.py` script cannot be run due to a syntax error (line 89 is not a comment, it opens a string variable, but never closes it):
https://github.com/fatiando/harmonica/blob/3c19d0c45fd014f1544f81bfed9eaf0e85ce2a32/harmonica/isostasy_pratt.py#L89

Try to reorganize the files as I recommended and then we can start getting into more detail with the code itself!
Don't hesitate to ask for help if you need it!

andieie:
I'm working on this in another branch and open a new pull request. I will note the changes there.

PR #203, opened 2020-11-04
Parallelize EQLs predictions and Jacobian build

santisoler:
Parallelize the build of the Jacobian matrix and the prediction step on
equivalent layer gridders through `numba.prange` iterators. This enables the
`fit` and `predict` methods to use all the available CPUs, greatly reducing the
computational time. The `prange` is only applied to the outer loops for each
task, i.e. distributing computation points between CPUs, but leaving the
iteration through the sources on single ones.

**Reminders**:

### DIFF
--- a/harmonica/equivalent_layer/utils.py
+++ b/harmonica/equivalent_layer/utils.py
@@ -5,19 +5,19 @@ from warnings import warn
 from numba import jit, prange
 
 
-@jit(nopython=True)
+@jit(nopython=True, parallel=True)
 def jacobian_numba(coordinates, points, jac, greens_function):
     """
     Calculate the Jacobian matrix using numba to speed things up.
 
-    It works both for Cartesian and spherical coordiantes.
-    We need to pass the corresponding Green's function through the
-    ``greens_function`` argument.
+    It works both for Cartesian and spherical coordiantes. We need to pass the
+    corresponding Green's function through the ``greens_function`` argument.
+    The Jacobian is built in parallel in order to reduce the computation time.
     """
     east, north, upward = coordinates[:]
     point_east, point_north, point_upward = points[:]
-    for i in range(east.size):
-        for j in range(point_east.size):
+    for i in prange(east.size):
+        for j in prange(point_east.size):
             jac[i, j] = greens_function(
                 east[i],
                 north[i],
--- a/harmonica/equivalent_layer/utils.py
+++ b/harmonica/equivalent_layer/utils.py
@@ -17,7 +17,7 @@ def jacobian_numba(coordinates, points, jac, greens_function):
     east, north, upward = coordinates[:]
     point_east, point_north, point_upward = points[:]
     for i in prange(east.size):
-        for j in prange(point_east.size):
+        for j in range(point_east.size):
             jac[i, j] = greens_function(
                 east[i],
                 north[i],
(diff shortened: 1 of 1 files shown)


leouieda (approved):
:+1: but we need to be careful when combing this with Dask to turn off the numbs parallelism. It would be interesting to have a Dask delayed version of  `grid` so we can turn this off for cross-validation but then still have parallel gridding.

santisoler:
Yes! I was thinking the same thing and wanted to test how Dask + Numba parallelization works. Maybe there's no need for switching on and off the Numba `parallel=True` and Dask assigns a single CPU to each Numba job :thinking:

santisoler:
I've just run this gridding notebook https://github.com/fatiando/transform2020-lightning/tree/master/santi with the `EQLHarmonic` from Harmonica v0.1.0 and with the parallel version from this PR. The cross validation (using dask) take the same time on both cases, while the prediction on a regular grid is obviously faster when parallel (more than half of the time). So I suspect that Dask might be smart enough to know how to handle Numba parallelizations. If anyone has better insight on this one, feel free to jump in!

leouieda:
Dask isn't smart enough but I think the problems are small enough not to show up. But what happens is that the CPU cores are over subscribed so we may not get the best performance possible. But we can turn off numba parallelism with an environment variable. So I think this is safe to merge. We can add more support for dask when we see problems arise.

harmonica/equivalent_layer/utils.py
```diff
@@ -28,19 +28,21 @@ def jacobian_numba(coordinates, points, jac, greens_function):
...
+    It works both for Cartesian and spherical coordiantes. We need to pass the
+    corresponding Green's function through the ``greens_function`` argument.
+    The prediction is run in parallel in order to reduce the computation time.
     """
     east, north, upward = coordinates[:]
     point_east, point_north, point_upward = points[:]
-    for i in range(east.size):
-        for j in range(point_east.size):
+    for i in prange(east.size):
+        for j in prange(point_east.size):
```
leouieda, line 45: Might be worth benchmarking if it's better to have a `prange` in the outer `for` only.

santisoler: I haven't seen any significant difference, but I think it's wiser to leave only `prange` in the outer loop only, we will never have that many cores to split the computation of almost one matrix element per core (remind me of reading this same sentence in 30 years from now haha).

santisoler:
> Dask isn't smart enough but I think the problems are small enough not to show up. But what happens is that the CPU cores are over subscribed so we may not get the best performance possible. But we can turn off numba parallelism with an environment variable. So I think this is safe to merge. We can add more support for dask when we see problems arise.

Ok. I don't have too much experience with parallelization and Dask is surprisingly difficult for me, probably I need to spend more time with it. So my thoughts are:
1. Consider first the regular user experience: most user will have several CPUs on their computers (even the low-profile computers nowdays have more than one core), so I would say that every CPU demanding task in Harmonica should leverage the available cores to reduce computation time without hassle. I think `prange` fills that gap.
2. We should also leave the possibility of more complex parallelization of the code through Dask. So if we can turn off Numba parallelization with an environment variable, I think the jitted functions won't need any modification to turn parallelization on and off.

leouieda:
👍🏽 I agree with both points 🙂

PR #205, opened 2020-11-06
Optimize forward models

### LINKED ISSUE #204: Parallelize forward models

Since #203 I found out that the EQL predictions were using a single CPU, achieving very high and annoying computation times.
After parallelizing the `predict` method, the computation time of the gridding significantly dropped.
I think we should also parallelize the forward models of **points**, **prisms** and **tesseroids** and benchmark them against their non-parallel versions.

One possible option is to use `prange` on jitted functions (enabling `parallel=True`), although maybe a Dask related solution would create more robust solutions that could be directly applied to distributed systems.

santisoler:
Parallelize the outer loops of `prism_gravity` and `point_mass_gravity`. Add an
optional `parallel` argument to `prism_gravity` and `point_mass_gravity` to run
them in parallel or in serial at will. Refactor and optimize tesseroids forward
model: avoid defining a set of point masses to compute their effect, instead
compute results in runtime. Simplify its code, reduce memory consumption and
prevent tesseroid forward model from using `jit_point_mass_spherical` function
which has been parallelized, but we don't want to use its parallel version when
computing GLQ on tesseroids forward model.

Fixes #204 

**Reminders**:

### DIFF
--- a/harmonica/equivalent_layer/utils.py
+++ b/harmonica/equivalent_layer/utils.py
@@ -12,9 +12,7 @@ from numba import jit, prange
 
 
 @jit(nopython=True, parallel=True)
-def jacobian_numba(
-    coordinates, points, jac, greens_function
-):  # pylint: disable=not-an-iterable
+def jacobian_numba(coordinates, points, jac, greens_function):
     """
     Calculate the Jacobian matrix using numba to speed things up.
 
--- a/harmonica/equivalent_layer/utils.py
+++ b/harmonica/equivalent_layer/utils.py
@@ -6,7 +6,9 @@ from numba import jit, prange
 
 
 @jit(nopython=True, parallel=True)
-def jacobian_numba(coordinates, points, jac, greens_function):
+def jacobian_numba(
+    coordinates, points, jac, greens_function
+):  # pylint: disable=not-an-iterable
     """
     Calculate the Jacobian matrix using numba to speed things up.
 
--- a/harmonica/forward/point_mass.py
+++ b/harmonica/forward/point_mass.py
@@ -240,11 +240,16 @@ def get_kernel(coordinate_system, field):
         "spherical": {
             "potential": kernel_potential_spherical,
             "g_z": kernel_g_z_spherical,
+            "g_northing": None,
+            "g_easting": None,
         },
     }
     if field not in kernels[coordinate_system]:
         raise ValueError("Gravitational field {} not recognized".format(field))
-    return kernels[coordinate_system][field]
+    kernel = kernels[coordinate_system][field]
+    if kernel is None:
+        raise NotImplementedError
+    return kernel
 
 
 @jit(nopython=True)
--- a/harmonica/forward/point_mass.py
+++ b/harmonica/forward/point_mass.py
@@ -2,7 +2,7 @@
 Forward modelling for point masses
 """
 import numpy as np
-from numba import jit
+from numba import jit, prange
 
 from ..constants import GRAVITATIONAL_CONST
 from .utils import check_coordinate_system, distance_cartesian, distance_spherical_core
@@ -214,7 +214,7 @@ def point_mass_gravity(
     return result.reshape(cast.shape)
 
 
-@jit(nopython=True)
+@jit(nopython=True, parallel=True)
 def jit_point_mass_cartesian(
     easting, northing, upward, easting_p, northing_p, upward_p, masses, out, kernel
 ):  # pylint: disable=invalid-name
@@ -237,7 +237,7 @@ def jit_point_mass_cartesian(
         Kernel function that will be used to compute the gravitational field on
         the computation points.
     """
-    for l in range(easting.size):
+    for l in prange(easting.size):
         for m in range(easting_p.size):
             out[l] += masses[m] * kernel(
                 easting[l],
@@ -305,7 +305,7 @@ def kernel_g_easting_cartesian(
     return -(easting - easting_p) / distance ** 3
 
 
-@jit(nopython=True)
+@jit(nopython=True, parallel=True)
 def jit_point_mass_spherical(
     longitude, latitude, radius, longitude_p, latitude_p, radius_p, masses, out, kernel
 ):  # pylint: disable=invalid-name
@@ -341,7 +341,7 @@ def jit_point_mass_spherical(
     cosphi_p = np.cos(latitude_p)
     sinphi_p = np.sin(latitude_p)
     # Compute gravitational field
-    for l in range(longitude.size):
+    for l in prange(longitude.size):
         for m in range(longitude_p.size):
             out[l] += masses[m] * kernel(
                 longitude[l],
--- a/harmonica/forward/point_mass.py
+++ b/harmonica/forward/point_mass.py
@@ -168,7 +168,7 @@ def point_mass_gravity(
         point masses.
         Available coordinates systems: ``cartesian``, ``spherical``.
         Default ``cartesian``.
-    parallel : bool
+    parallel : bool (optional)
         If True the computations will run in parallel using Numba built-in
         parallelization. If False, the forward model will run on a single core.
         Might be useful to disable parallelization if the forward model is run
--- a/harmonica/forward/point_mass.py
+++ b/harmonica/forward/point_mass.py
@@ -9,7 +9,13 @@ from .utils import check_coordinate_system, distance_cartesian, distance_spheric
 
 
 def point_mass_gravity(
-    coordinates, points, masses, field, coordinate_system="cartesian", dtype="float64"
+    coordinates,
+    points,
+    masses,
+    field,
+    coordinate_system="cartesian",
+    parallel=True,
+    dtype="float64",
 ):
     r"""
     Compute gravitational fields of point masses.
@@ -156,6 +162,11 @@ def point_mass_gravity(
         point masses.
         Available coordinates systems: ``cartesian``, ``spherical``.
         Default ``cartesian``.
+    parallel : bool
+        If True the computations will run in parallel using Numba built-in
+        parallelization. If False, the forward model will run on a single core.
+        Might be useful to disable parallelization if the forward model is run
+        by an already parallelized workflow. Default to True.
     dtype : data-type (optional)
         Data type assigned to resulting gravitational field. Default to
         ``np.float64``.
@@ -170,10 +181,16 @@ def point_mass_gravity(
         Marussi tensor components in Eotvos.
     """
     # Organize dispatchers and kernel functions inside dictionaries
-    dispatchers = {
-        "cartesian": jit_point_mass_cartesian,
-        "spherical": jit_point_mass_spherical,
-    }
+    if parallel:
+        dispatchers = {
+            "cartesian": jit_point_mass_cartesian_parallel,
+            "spherical": jit_point_mass_spherical_parallel,
+        }
+    else:
+        dispatchers = {
+            "cartesian": jit_point_mass_cartesian,
+            "spherical": jit_point_mass_spherical,
+        }
     kernels = {
         "cartesian": {
             "potential": kernel_potential_cartesian,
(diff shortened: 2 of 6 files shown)


santisoler:
Tests run much faster with these optimizations (from ~70s to ~20s), specially the tesseroid related ones due to the removal of the `parallel=True` argument on the `jit` decorator. Parallelization of tesseroids is something I want to implement, but the `stack` and `small_tesseroids` arrays are allocated only once for every iterations. In order to parallelize the whole process, we need to be sure that each CPU gets its own `stack` and `small_tesseroids` array, otherwise the results will be wrong.

leouieda (commented):
@santisoler I'm not a fan of returning the tesseroid code to the way it was. And I think this is hinting at a larger issue.

Our forward modelling functions are a bit too complex. This is because they are trying to serve two purposes:

1. Be friendly to end users doing terrain corrections, etc.
2. Allow composition by people developing inversions, etc, with them.

I think those are two separate and conflicting goals for these functions. So my proposal is to make 2 functions:

1. Low-level: just calculates the fields for **one** element on the given computation points. No checks, no extra work. Maybe do the broadcasting but I would be fine with not doing it. These are meant for developers.
2. High-level: uses the low level functions to accomplish tasks the user requires (terrain correction, forward modelling with checks and parallelism, etc). 

The parallelism in 2 could be done with dask or with numba (making a jit function that calls the low-level functions). The low-level could even be pure `numba.jit` functions for simplicity.

And before you say it, looping over model elements outside of numba doesn't make a difference 😉 https://gist.github.com/leouieda/94b81026d57229f30ccb3bddfc1892e7 At least not enough to make it worth the hassle. 

Personally, I think we should focus on implementing the low-level functions only for now. We can work on the high-level once we have a good use case to benchmark and design (like terrain correction with separation of near and far zone).

What do you think?

leouieda:
Also, we need to have an option without parallelism as the core. Otherwise, these functions are hard to use for building Jacobians.

leouieda:
I've been doing some investigation and learned a few things:

* Using `prange` over the computation points and looping over model elements in Python works and provides decent speed-ups. 
* Parallelizing the model element loop with Dask is harder since we'd need to split the model domain instead of just dispatching jobs for every single model element. Also, this would involving memory allocation since we can't have multiple jobs writing to the same array.
* We can create jit functions programmatically by not using `numba.jit` as a decorator:  `prism_jit = numba.jit(nopython=True, parallel=True)(prism_func)`. We can have `prange` in the function and turn off parallelism by passing `parallel=False`.

santisoler:
Thanks for the insight @leouieda!
I was wondering if the last point you notice was possible, thanks Python for programmatic decorations :rocket:
I absolutely agree that we should let users to enable/disable parallelism. Sometimes is really useful to have it (forward modelling, single gridder predictions), but when the forward model is incorporated on an already parallelized workflow, maybe we want to use a single core for each one (cross-validation for example).

Regarding the previous comments, I wanted to discuss them personally so we can understand better, but no hurries, I think I have enough homework for the week so I don't think I'll get into this in a few days.

leouieda:
We can chat about this at the Fatiando meeting today.

harmonica/forward/tesseroid.py
```diff
@@ -257,131 +252,142 @@ def jit_tesseroid_gravity(
...
+    result = 0.0
+    for i, lon_node in enumerate(lon_nodes):
+        for j, lat_node in enumerate(lat_nodes):
+            for k, rad_node in enumerate(rad_nodes):
+                # Get coordinates of the point mass
+                longitude_p = np.radians(0.5 * (e - w) * lon_node + 0.5 * (e + w))
+                latitude_p = np.radians(0.5 * (n - s) * lat_node + 0.5 * (n + s))
+                radius_p = 0.5 * (top - bottom) * rad_node + 0.5 * (top + bottom)
+                cosphi_p = np.cos(latitude_p)
+                sinphi_p = np.sin(latitude_p)
```
santisoler, line 363: Move the definition of some of these variables one or two loops out.

santisoler:
@leouieda I've made some changes on the tesseroid and the point mass forward models.

On the tesseroids I've changed the order of the loops and moved the definition of some point mass coordinates so they don't get recomputed without need.

On the point masses tried to add the `parallel` flag following these criteria:
- the parallel and no-parallel jitted functions should be compiled only once, even if the `point_mass_gravity` function is run more than once.
- The function should work with `NUMBA_JIT_DISABLED=1` (that's why the `if`s statements on the bottom of the file.

Let me know what do you think or if you have though of something better.

harmonica/forward/point_mass.py
```diff
@@ -220,39 +215,47 @@ def point_mass_gravity(
...
+        },
+        "spherical": {
+            "potential": kernel_potential_spherical,
+            "g_z": kernel_g_z_spherical,
+            "g_northing": None,
+            "g_easting": None,
+        },
+    }
+    if field not in kernels[coordinate_system]:
+        raise ValueError("Gravitational field {} not recognized".format(field))
```
santisoler, line 254:
```suggestion
        raise ValueError("Gravitational field '{}' not recognized".format(field))
```

harmonica/forward/tesseroid.py
```diff
@@ -263,131 +258,148 @@ def jit_tesseroid_gravity(
...
-                coordinates[1][m : m + 1],  # slice lat to pass a single element array
-                coordinates[2][m : m + 1],  # slice rad to pass a single element array
-                point_masses[0, :n_point_masses],
-                point_masses[1, :n_point_masses],
-                point_masses[2, :n_point_masses],
-                density[l] * weights[:n_point_masses],
-                result[m : m + 1],
-                kernel,
-            )
+            # Compute effect of the tesseroid though GLQ
```
santisoler, line 285:
```suggestion
            # Compute effect of the tesseroid through GLQ
```

harmonica/forward/tesseroid.py
```diff
@@ -263,131 +258,148 @@ def jit_tesseroid_gravity(
...
+        sinphi_p = np.sin(latitude_p)
+        for k, rad_node in enumerate(rad_nodes):
+            # Get the radii of the point masses
+            radius_p = 0.5 * (top - bottom) * rad_node + 0.5 * (top + bottom)
+            # Get kappa constant and the mass of the point mass
+            kappa = radius_p ** 2 * cosphi_p
+            for i, lon_node in enumerate(lon_nodes):
+                # Get the longitudes of the point masses
+                longitude_p = np.radians(0.5 * (e - w) * lon_node + 0.5 * (e + w))
+                mass = (
```
santisoler, line 378:
```suggestion
                # Compute the mass of the point masses
                mass = (
```

harmonica/forward/tesseroid.py
```diff
@@ -263,131 +258,148 @@ def jit_tesseroid_gravity(
...
+        latitude_p = np.radians(0.5 * (n - s) * lat_node + 0.5 * (n + s))
+        cosphi_p = np.cos(latitude_p)
+        sinphi_p = np.sin(latitude_p)
+        for k, rad_node in enumerate(rad_nodes):
+            # Get the radii of the point masses
+            radius_p = 0.5 * (top - bottom) * rad_node + 0.5 * (top + bottom)
+            # Get kappa constant and the mass of the point mass
+            kappa = radius_p ** 2 * cosphi_p
+            for i, lon_node in enumerate(lon_nodes):
+                # Get the longitudes of the point masses
```
santisoler, line 376:
```suggestion
                # Get the longitudes of the point mass
```

harmonica/forward/tesseroid.py
```diff
@@ -263,131 +258,148 @@ def jit_tesseroid_gravity(
...
+    result = 0.0
+    for j, lat_node in enumerate(lat_nodes):
+        # Get the latitudes of the point masses
+        latitude_p = np.radians(0.5 * (n - s) * lat_node + 0.5 * (n + s))
+        cosphi_p = np.cos(latitude_p)
+        sinphi_p = np.sin(latitude_p)
+        for k, rad_node in enumerate(rad_nodes):
+            # Get the radii of the point masses
+            radius_p = 0.5 * (top - bottom) * rad_node + 0.5 * (top + bottom)
+            # Get kappa constant and the mass of the point mass
```
santisoler, line 373:
```suggestion
            # Get kappa constant for the point mass
```

harmonica/forward/tesseroid.py
```diff
@@ -263,131 +258,148 @@ def jit_tesseroid_gravity(
...
+    # optimization: reduce the number of times that the trigonometric functions
+    # are evaluated)
+    result = 0.0
+    for j, lat_node in enumerate(lat_nodes):
+        # Get the latitudes of the point masses
+        latitude_p = np.radians(0.5 * (n - s) * lat_node + 0.5 * (n + s))
+        cosphi_p = np.cos(latitude_p)
+        sinphi_p = np.sin(latitude_p)
+        for k, rad_node in enumerate(rad_nodes):
+            # Get the radii of the point masses
```
santisoler, line 371:
```suggestion
            # Get the radii of the point mass
```

harmonica/forward/tesseroid.py
```diff
@@ -263,131 +258,148 @@ def jit_tesseroid_gravity(
...
-                    )
-                    mass_index += 1
+    # Compute effect of the tesseroid on the observation point
+    # by iterating over the location of the point masses
+    # (move the iteration along the longitudinal nodes to the bottom for
+    # optimization: reduce the number of times that the trigonometric functions
+    # are evaluated)
+    result = 0.0
+    for j, lat_node in enumerate(lat_nodes):
+        # Get the latitudes of the point masses
```
santisoler, line 366:
```suggestion
        # Get the latitudes of the point mass
```

harmonica/forward/tesseroid.py
```diff
@@ -263,131 +258,149 @@ def jit_tesseroid_gravity(
...
-                    )
-                    mass_index += 1
+    # Compute effect of the tesseroid on the observation point
+    # by iterating over the location of the point masses
+    # (move the iteration along the longitudinal nodes to the bottom for
+    # optimization: reduce the number of times that the trigonometric functions
+    # are evaluated)
+    result = 0.0
+    for j, lat_node in enumerate(lat_nodes):
+        # Get the latitudes of the point mass
```
santisoler, line 366:
```suggestion
        # Get the latitude of the point mass
```

harmonica/forward/tesseroid.py
```diff
@@ -263,131 +258,149 @@ def jit_tesseroid_gravity(
...
+        latitude_p = np.radians(0.5 * (n - s) * lat_node + 0.5 * (n + s))
+        cosphi_p = np.cos(latitude_p)
+        sinphi_p = np.sin(latitude_p)
+        for k, rad_node in enumerate(rad_nodes):
+            # Get the radii of the point mass
```
santisoler, line 371:
```suggestion
            # Get the radius of the point mass
```

harmonica/forward/tesseroid.py
```diff
@@ -263,131 +258,149 @@ def jit_tesseroid_gravity(
...
+            radius_p = 0.5 * (top - bottom) * rad_node + 0.5 * (top + bottom)
+            # Get kappa constant for the point mass
+            kappa = radius_p ** 2 * cosphi_p
+            for i, lon_node in enumerate(lon_nodes):
+                # Get the longitudes of the point mass
```
santisoler, line 376:
```suggestion
                # Get the longitude of the point mass
```

harmonica/equivalent_layer/utils.py
```diff
@@ -12,7 +12,9 @@
 
 
 @jit(nopython=True, parallel=True)
-def jacobian_numba(coordinates, points, jac, greens_function):
+def jacobian_numba(
+    coordinates, points, jac, greens_function
+):  # pylint: disable=not-an-iterable
```
santisoler, line 17: This is not related to the optimization of forward models, and it will conflict with the changes introduced on #207
```suggestion
```

PR #227, opened 2021-04-14
Use the Jupyter book sphinx theme instead of RTD

leouieda:
The theme is a lot nicer and more modern. It has builtin a lot of the hacks we made to the template to add download and edit buttons. We could also do away with most of our custom templating and CSS tweaks. This opens up the possibility for further style improvements in the future (using admonitions, panels, etc). 

Notable changes: 
* the version number is now included in the page title
* the banner includes the slogan
* the front page disclaimer is redesigned and not gotten from the README anymore
* external links are included in `index.rst` instead of `conf.py` (this means no icons and different titles but still looks nice)
* Copyright notice only includes the current year since it applies to the current version

Related to fatiando/maintenance#13

**Reminders**:


santisoler:
Nice! It looks way cleaner. Does it have a dark mode? :eyes: haha

> Check if the google analytics code is inserted

I've served the docs locally and checked that the Google analytics script is loaded in the `head`. Do we _really_ need it? I would remove it unless we have a very strong reason to keep it.

leouieda:
> Does it have a dark mode?

I'm sure they'd love a PR adding that 😉 

> Do we really need it? I would remove it unless we have a very strong reason to keep it.

I kept it since that is one of the few indicators we have of how many people find and use the projects. The download numbers are kind of meaningless with CI and citations are sparse. If we ever go for grant funding, this is a number that should include. The compromise is that we don't gather IP data.

> Check links to edit sphinx gallery content

The edit button for sphinx-gallery output is broken since the sources aren't really in the repo. This is something that would best added to the original template rather than us baking our own custom solution. So I'm fine with it being broken for now (nobody ever used those links anyway). What do you think?

Other than that, I think this is done. We can start using the new style in the docs now and improving the layout in subsequent PRs. This was surprisingly smooth and not much work at all 🙂

santisoler (approved):
> I kept it since that is one of the few indicators we have of how many people find and use the projects. The download numbers are kind of meaningless with CI and citations are sparse. If we ever go for grant funding, this is a number that should include. The compromise is that we don't gather IP data.

Ok, I get it. We can discuss this later, but another step we could do is to try to move away from Google Analytics to a more transparent data gathering tool.

> The edit button for sphinx-gallery output is broken since the sources aren't really in the repo. This is something that would best added to the original template rather than us baking our own custom solution. So I'm fine with it being broken for now (nobody ever used those links anyway). What do you think?

I agree. If anyone is willing to solve a problem with one of our examples, I expect they would find the Contributing guidelines and open an issue or PR to solve it.

I think this is ready to go. I just left a minor comment related to the copyright notice.
After that, I think this is ready to be merged.

doc/conf.py
```diff
@@ -4,16 +4,25 @@
...
-from sphinx_gallery.sorting import FileNameSortKey
+from sphinx_gallery.sorting import ExampleTitleSortKey
 
 from harmonica.version import full_version
 
+
+# Project information
+# -----------------------------------------------------------------------------
+project = "Harmonica"
+copyright = f"{datetime.date.today().year}, The {project} Developers"
```
leouieda, line 18:
```suggestion
copyright = f"2018-{datetime.date.today().year}, The {project} Developers"
```

Ah, right. Sorry forgot about that. Fixed!

leouieda:
>  We can discuss this later, but another step we could do is to try to move away from Google Analytics to a more transparent data gathering tool.

I've been searching for alternatives but couldn't find anything that was free or didn't require setting up our own server. If you find anything, let me know!

> I think this is ready to go. I just left a minor comment related to the copyright notice. After that, I think this is ready to be merged.

Fixed and merging now!

PR #236, opened 2021-05-14
Add option to set EQLHarmonic points to constant depth

santisoler:
Rename the `relative_depth` parameter for `depth` and add a new `depth_type`
parameter. Keep supporting the `relative_depth` parameter but raise
a FutureWarning if passed. Add a new `build_points()` method for creating
the point source coordinates based on the data points.

**Reminders**:

### DIFF
--- a/harmonica/equivalent_layer/harmonic.py
+++ b/harmonica/equivalent_layer/harmonic.py
@@ -74,13 +74,27 @@ class EQLHarmonic(vdb.BaseGridder):
         If None, will place one point source bellow each observation point at
         a fixed relative depth bellow the observation point [Cooper2000]_.
         Defaults to None.
-    relative_depth : float
-        Relative depth at which the point sources are placed beneath the
+    depth : float
+        Parameter used to control the depth at which the point sources will be
+        located.
+        If ``depth_type`` is equal to ``"relative"``, the ``depth`` specifies
+        the relative depth at which the point sources are placed beneath the
         observation points. Each source point will be set beneath each data
         point at a depth calculated as the elevation of the data point minus
-        this constant *relative_depth*. Use positive numbers (negative numbers
-        would mean point sources are above the data points). Ignored if
-        *points* is specified.
+        this *depth*. Use positive numbers (negative numbers would mean point
+        sources are above the data points).
+        If ``depth_type`` is equal to ``"constant"``, the ``depth`` specifies
+        the constant depth at which the point sources are placed beneath the
+        observation points. Every source point will be located at this *depth*.
+        Use positive numbers (negative numbers would mean point sources are
+        located above the zeroth level).
+        This parameter is ignored if *points* is specified.
+        Defaults to 500.
+    depth_type : str
+        Strategy used for setting the depth of the point sources.
+        The two available strategies are ``"constant"`` and ``"relative"``.
+        This parameter is ignored if *points* is *points* is specified.
+        Defaults to ``"relative"``.
     parallel : bool
         If True any predictions and Jacobian building is carried out in
         parallel through Numba's ``jit.prange``, reducing the computation time.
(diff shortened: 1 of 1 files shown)


harmonica/equivalent_layer/harmonic.py
```diff
@@ -155,17 +187,54 @@ def fit(self, coordinates, data, weights=None):
...
-                coordinates[2] - self.relative_depth,
-            )
+            self.points_ = self.build_points(coordinates)
         else:
             self.points_ = vdb.n_1d_arrays(self.points, 3)
         jacobian = self.jacobian(coordinates, self.points_)
         self.coefs_ = vdb.least_squares(jacobian, data, weights, self.damping)
         return self
 
+    def build_points(self, coordinates):
```
leouieda, line 197: We probably want to make this private at first. If there is a need, we can always make it public later.

harmonica/equivalent_layer/harmonic.py
```diff
@@ -74,13 +75,27 @@ class EQLHarmonic(vdb.BaseGridder):
...
-        *points* is specified.
+        this *depth*. Use positive numbers (negative numbers would mean point
+        sources are above the data points).
+        If ``depth_type`` is equal to ``"constant"``, the ``depth`` specifies
+        the constant depth at which the point sources are placed beneath the
+        observation points. Every source point will be located at this *depth*.
+        Use positive numbers (negative numbers would mean point sources are
+        located above the zeroth level).
+        This parameter is ignored if *points* is specified.
+        Defaults to 500.
```
leouieda, line 93: It's so hard to choose a sensible default here. This is OK for now but we should try to think of a way to define this relative to some property of the input data (average spacing, etc).

harmonica/equivalent_layer/harmonic.py
```diff
@@ -113,15 +128,32 @@ def __init__(
...
-        self.relative_depth = relative_depth
+        self.depth = depth
+        self.depth_type = depth_type
         self.parallel = parallel
         # Define Green's function for Cartesian coordinates
         self.greens_function = greens_func_cartesian
+        # Check if depth_type is valid
+        if depth_type not in ("constant", "relative"):
+            raise ValueError(
+                "Invalid depth_type '{}'. It can either be 'constant' or 'relative'."
```
leouieda, line 146:
```suggestion
               f"Invalid depth type '{depth_type}'. Should be either be 'constant' or 'relative'."
```

harmonica/tests/test_eql_harmonic.py
```diff
@@ -129,6 +130,89 @@ def test_eql_harmonic_small_data_cartesian():
...
+    with warnings.catch_warnings(record=True) as warn:
+        eql = EQLHarmonic(relative_depth=4.5e3)
+        assert len(warn) == 1
+        assert issubclass(warn[-1].category, FutureWarning)
+    npt.assert_allclose(eql.depth, 4.5e3)
+    assert eql.depth_type == "relative"
+    points = eql._build_points(coordinates)
+    # Check output
+    expected_points = (easting, northing, upward - 4.5e3)
+    npt.assert_allclose(points, expected_points)
```
leouieda, line 167: Best to split these tests into multiple ones. Use a fixture for the data and parametrize the options. Otherwise it's hard to tell which part breaks when tests fail.

leouieda (commented):
Just a small suggestion to parametrize the test functions. Otherwise looking really good!

harmonica/equivalent_layer/harmonic.py
santisoler: I agree that it's very hard to choose a default value for the `depth`. Nevertheless I've always saw it as a dummy value that people are encouraged to change and experiment with. I'm afraid to set a clever way to get it and people think that our  automated choice would be smarter than theirs.

harmonica/tests/test_eql_harmonic.py
santisoler: Sure!

santisoler:
I've improved the tests by defining a `coordinates` fixture and by parametrizing the depth_type and the expected `upward` array.
I created a new test function for the `relative_depth` parameters. It will make it easier to deprecate it in the future.

PR #237, opened 2021-05-14
Generate version string on _version.py on build

santisoler:
Move `setuptools-scm` to `requirements-dev.txt`. Automatically create
the version string on build and store it in `harmonica/_version.py`.
Remove the need to run `pgk_resources` to get the version number. Invert
`MANIFEST.in` by specifying what should be ignored because `setuptools-scm`
includes everything by default.

Related to fatiando/maintenance#4

Inspired on fatiando/pooch#235

**Reminders**:


santisoler:
@leouieda We are still using `pkg_resources` on `datasets/sample_data.py` to load the registry file. Should we leave that as it is or we would like to change that too?

leouieda:
@santisoler we would still need it for the registry so leave it as is. I'm leaning towards defining the registry as a Python dict in the module instead of the external file. It's no harder to do and easier to maintain since there is no messing with `setup.py` and `MANIFEST.in`.

leouieda (approved):
👍🏽  now we just pray that this doesn't break everything. Would be good to download and inspect the test PyPI package after this is merged.

PR #238, opened 2021-05-31
Compute upward derivative of a grid in the frequency domain

santisoler:
Define a new `derivative_upward` function for computing the spatial upward
derivative of a 2D grid in the frequency domain. The function makes use of
`xrft` for handling Fourier transformations of `xarray` objects. Add a new
`filters` subpackage that includes FFT filters: functions that take grids in
frequency domain and return the desired filter also in frequency domain. Add
`fft` and `ifft` wrapper functions of the `xrft.fft` and `xrft.ifft` ones. Add
a new `apply_filter` function that takes a grid in the spatial domain, applies
fft, the filter and ifft and returns the filtered grid also in spatial domain.
Add tests for the new features and a gallery example for the upward derivative.
Add `netcdf4` as requirement for testing.

**Reminders**:


RichardScottOZ:
Seems good - is 'vertical' terminology more used than upward?

Notes for the function could I suppose include parenthetical references to x/y after easting and northing?

santisoler:
Thanks @RichardScottOZ for the feedback!

> Seems good - is 'vertical' terminology more used than upward?

"Vertical" is more widely used, although the drawback of using it is that it doesn't explicit specify the direction of the "vertical" axe. That's why in Harmonica we decided to use "upward" for specifying the _non-horizontal_ direction explicitly stating that it points up. 

> Notes for the function could I suppose include parenthetical references to x/y after easting and northing?

And a similar thing happens with x and y. We decided to use "easting" and "northing" for the horizontal axes, this way we avoid the need to constantly remember where are the "x" and "y" axes pointing (some software use "x" for northing, while some other uses for the easting).

Do you feel that using `easting`, `northing` and `upward` might be confusing?

ThomasMGeo:
I would just put in the comments that upward is also called vertical, but I do like your reasoning of calling it upward.

taschaap:
These are great! Might I suggest adding an analytic signal function to the bunch?

santisoler:
> I would just put in the comments that upward is also called vertical, but I do like your reasoning of calling it upward.

@ThomasMGeo do you mean in the comments or in the docstrings?

ThomasMGeo:
I think either would work. Maybe just the comments to not clutter the doc strings?

santisoler:
> I think either would work. Maybe just the comments to not clutter the doc strings?

I agree. I rather avoid using `vertical` in the docstring in order to avoid multiple ways of referring the same thing. So no problem of adding it to the comments.

santisoler:
> tutorial, yes, good idea - do you have a barebones run this? For horizontal array derivatives for magnetics or something - be happy to write one up if you do

Thanks for the initiative @RichardScottOZ! I don't have any, but it might be a good use case for the magentic data samples that you've been proposing on Slack. My idea for this tutorial would be to show how we can compute the horizontal derivatives of a grid through finite differences with `xarray` and how we could use the future `harmonica.derivative_upward` for computing the vertical one.

santisoler:
I put some work on this PR so we can merge it and implement all the FFT filters that @LL-Geo wrote on #299 .

I added more tests (with a sample gravity grid) and include padding both in tests and in the example.
I haven't worked too much on the gallery example since I'm probably going to refactor it on #305 .
Would love to have additional eyes on this one @leouieda @LL-Geo @RichardScottOZ .

santisoler:
I think this is ready as a first design for FFT transformations and functions. If we would like to change anything we can do it in the future, no big deal. After merging, next step would be to incorporate the FFT filters that @LL-Geo defined on #299 into this framework of filters + transformation functions.

I'm ignoring a minor CI error due to a small difference in tesseroids doctest. That issue is completely unrelated to this PR.

PR #240, opened 2021-08-27
Extend support for Python 3.10

santisoler:
Bump the latest end on GitHub Actions tests to Python 3.10. Build docs in
GitHub Actions using Python 3.10. Bump environment to Python 3.10. Add Python
3.10 to setup.cfg.

Related to fatiando/maintenance#15


hugovk:
For pooch and boule, 3.10-dev is needed because it's passed to https://github.com/actions/setup-python to set up the Python version.

However, here we're passing it to https://github.com/conda-incubator/setup-miniconda so we'll need to check if they have a version of 3.10 available and how to use it.

santisoler:
Good news: now tests pass on Python 3.10! :tada:

Bad news: there's a compatibility problem between Numba and Numpy on Python 3.10 when installed through conda/mamba.
Numba 0.55.* needs Numpy <= 1.21 because Numpy 1.22 is not supported yet (see https://github.com/numba/numba/issues/7754). Therefore, right now we cannot install Harmonica under Python 3.10 using conda or mamba, and so the docs fail to build.

Maybe we should wait until Numba add further support, or should we pin Numpy for now?

cc @leouieda @LL-Geo

leouieda:
Great! I would wait for this a bit more since there's no point adding a pinning which we'll have to remove soon anyway.

hugovk:
```
ERROR: Could not find a version that satisfies the requirement vtk>=9 (from versions: none)
ERROR: No matching distribution found for vtk>=9
```

No Python 3.10 wheels yet: https://pypi.org/project/vtk/#files

https://gitlab.kitware.com/vtk/vtk/-/issues/18335 is the tracking issue.

A fix has been merged: https://gitlab.kitware.com/vtk/vtk/-/issues/18335#note_1164493 but not released.

For a workaround there's an unofficial build: https://gitlab.kitware.com/vtk/vtk/-/issues/18335#note_1180656 

See https://discourse.vtk.org/t/vtk-9-2-0-release-cycle/8149 for planned VTK release schedule, but I don't see any of the planned release candidates at https://pypi.org/project/vtk/#history so I guess they're running a bit behind.

santisoler:
Thanks for the hint @hugovk.

That workaround is an unofficial build made by pyvista. In fact, in `conda-forge` we can install VTK on Python 3.10: https://github.com/pyvista/pyvista/discussions/2064

So, the things that are holding us back are:
- No official build of VTK for Python 3.10 in PyPI (solvable through the unofficial build)
- Numba needs Numpy <= 1.21 (we can pin Numpy for now)

Should we keep waiting or apply those patches now **and remember to remove them in the future**?

cc @leouieda

santisoler:
After more time than I wanted, I think we can safely extend support to Python 3.10. The tests that are failing are only for Python 3.7 (see #355).

PR #242, opened 2021-08-30
Pass density as numpy array on prism_layer example

santisoler:
Patch the `prism_layer` example so the density of the prisms is passed as
a Numpy array instead as a `DataArray`. This solves an `xarray` error when
building the docs. This is just a patch, although the problem must be solved
upstream in Verde.

Related to https://github.com/fatiando/verde/issues/330

**Reminders**:


santisoler:
This PR is just a patch, although the problem must be solved upstream in Verde.

santisoler:
This PR is a cheap solution that is not solving anything in the end. I'm closing it.

santisoler:
See #243 for a good candidate of a proper solution to the issue

PR #245, opened 2021-09-08
Transform require_numba decorator into a global variable

santisoler:
The require_numba as a function wrapper makes it hard to use
`pytest.mark.parametrize` on test functions. Following pytest
documentation, it's better to define custom decorators as global
variables.

Source: https://docs.pytest.org/en/6.2.x/skipping.html#id1


leouieda (approved):
👍🏽 good to know how to do this properly! Should have read the docs before making the decorator wrapper 😆

PR #255, opened 2021-10-18
Rename equivalent sources classes

### LINKED ISSUE #250: Rename EQLHarmonic and EQLHarmonicSpherical classes into something more meaningful

I would like to rename the equivalent layer/sources classes to something more meaningful.
Since both classes work with point sources rather an actual layer of mass distribution, I would be more prone to include "equivalent sources" in their name.
On the other hand, I don't think we should specify the "harmonic" nature of the fields these classes generate.
We don't have another type of equivalent sources interpolations, so I think we should not include the "harmonic" part until we really need someway to differentiate them from others (maybe we should differentiate future classes from these two instead).

I would like to add some backward compatibility on this, specially after the Harmonica tutorial at Transform21.
Maybe we could create two new child classes (whose names are `EQLHarmonic` and `EQLHarmonicSpherical`) that raises a FutureWarning after initialization.

Proposals:
- `EquivalentSources` and `EquivalentSourcesSpherical`
- `EqSources` and `EqSourcesSpherical`

santisoler:
Rename `EQLHarmonic` and `EQLHarmonicSpherical` classes to `EquivalentSources`
and `EquivalentSourcesSph`, respectively. Refactor files organization: rename
`equivalent_layer` directory to `equivalent_sources` and the sources files to
`cartesian.py` and `spherical.py`. Split the file with the tests function by
coordinate system: Cartesian and spherical. Rename some tests functions. Add
new child classes `EQLHarmonic` and `EQLHarmonicSph` for backwards
compatibility that raise a `FutureWarning` when initialized. Update gallery
examples and documentation pages to use "equivalent sources" instead of
"equivalent layer". Mention "equivalent layer" in one gallery example though.
Rename `eql` variables in examples and tests with `eqs`.

TODO:
- [x] Update examples
- [x] Update "Equivalent Layer" in docs
- [x] Rename `eql` variables in the examples and tests
- [x] Mention "equivalent layer" somewhere in the docs

Fixes #250

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,7 +38,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu, macos, windows]
-        python: [3.6, 3.7, 3.8, 3.9]
+        python: [3.6, 3.9]
         # If "optional", will install non-required dependencies in the build
         # environment. Otherwise, only required dependencies are installed.
         dependencies: [""]


leouieda:
That's a good point about the search terms. It may be good to leave it in a tutorial or the docstring as something like "equivalent sources (a.k.a. equivalent layer)" or something like that.

>I do have some doubts: we used to use eql for naming the equivalent layer instances on examples, would be better to rename then to equivalent_sources or eq_sources? 

How about `eqs` since we call a lot of methods with lots of arguments on it?

PR #275, opened 2021-10-27
Add gradient-boosted equivalent sources

santisoler:
Add a new `EquivalentSourcesGB` class that implements gradient-boosting
equivalent sources. It's a subclass of `EquivalentSources`, takes new
`window_size` and `random_state` arguments and has a modified `fit` method that
applies the gradient-boosting algorithm. The amount of overlapping is fixed to
50%. The windows are randomly shuffled by default. Add new test functions for
the new class and its features. Add a gallery example for gradient-boosted
equivalent sources.

**Reminders**:


harmonica/tests/test_gradient_boosted_eqs.py
```diff
@@ -0,0 +1,239 @@
...
+    # The interpolation should be sufficiently accurate on the data points
+    eqs = EquivalentSourcesGB(window_size=100e3)
+    eqs.fit(coordinates, data)
+    assert_mse(data, eqs.predict(coordinates), tol=1e-5 * vd.maxabs(data))
+
+    # Gridding onto a denser grid should be reasonably accurate when compared
+    # to synthetic values
+    grid = vd.grid_coordinates(region=region, shape=(60, 60), extra_coords=0)
+    true = hm.point_mass_gravity(grid, points, masses, field="g_z")
+    assert_mse(true, eqs.predict(grid), tol=1e-3 * vd.maxabs(data))
```
santisoler, line 165: Pay special attention to this!

harmonica/tests/test_gradient_boosted_eqs.py
```diff
@@ -0,0 +1,235 @@
...
+    # The interpolation should be sufficiently accurate on the data points
+    eqs = EquivalentSourcesGB(window_size=1e3, depth=1e3, damping=None, random_state=42)
+    eqs.fit(coordinates, data)
+    npt.assert_allclose(data, eqs.predict(coordinates), atol=1e-2 * vd.maxabs(data))
+
+    # Gridding onto a denser grid should be reasonably accurate when compared
+    # to synthetic values
+    grid = vd.grid_coordinates(region=region, shape=(60, 60), extra_coords=0)
+    true = hm.point_mass_gravity(grid, points, masses, field="g_z")
+    npt.assert_allclose(true, eqs.predict(grid), atol=1e-2 * vd.maxabs(true))
```
santisoler, line 161: @leouieda pay attention to this test function. Does it make sense to compare with `atol`?

harmonica/equivalent_sources/gradient_boosted.py
```diff
@@ -0,0 +1,382 @@
...
+            )
+            # Update the residue
+            residue -= predicted
+            # Add RMS of the residue to the RMSE
+            errors.append(np.sqrt(np.mean(residue ** 2)))
+            # Update source coefficients
+            self.coefs_[point_window] += coeffs_chunk
+        self.errors_ = np.array(errors)
+
+    def _create_windows(self, coordinates, shuffle_windows=True):
```
santisoler, line 302: I've added the `shuffle_windows` argument to this function in order to run a test that depends on not shuffling (the one that checks that the correspondence between sources windows and data windows). I'm not planning to make it public since it's a bad idea not to shuffle.

harmonica/tests/test_gradient_boosted_eqs.py
santisoler, line 161: @leouieda pay attention to this test function. Does it make sense to compare with atol?

examples/equivalent_sources/gradient_boosted.py
```diff
@@ -0,0 +1,138 @@
...
+the sources and data points that fall under each window at a time. The order in
+which the windows are visited is randomized to improve convergence of the
+algorithm.
+
+Here we will produce a grid out of a portion of the ground gravity survey from
+South Africa (see :func:`harmonica.datasets.fetch_south_africa_gravity`) using
+the gradient-boosted equivalent sources. This particlar dataset is not very
+large, in fact we could use the :class:`harmonica.EquivalentSources` instead.
+But we will use the :class:`harmonica.EquivalentSourcesGB` for illustrating how
+to use them on a simple example.
```
leouieda, line 28:
```suggestion
to use them on a small example.
```

examples/equivalent_sources/gradient_boosted.py
```diff
@@ -0,0 +1,138 @@
...
+
+# Start building the equivalent sources.
+# We'll use the block-averaged sources with a block size of 2km and windows of
+# 100km x 100km.
+window_size = 100e3
+block_size = 2e3
+
+# Let's estimate the memory required to store the largest Jacobian when using
+# these values for the window_size and the block_size.
+jacobian_req_memory = hm.EquivalentSourcesGB.estimate_required_memory(
```
leouieda, line 70: How about making this a regular method instead of a class method? Creating instances is cheap since the constructors don't do any computation. Then we could:

```python
eqs = hm.EquivalentSourcesGB(window_size=10, block_size=1)
print(eqs.estimate_required_memory(coordinates))
# Happy = carry on
# Too much = reduce window size and run again
```

This way you don't need to pass the arguments twice.

harmonica/equivalent_sources/gradient_boosted.py
```diff
@@ -0,0 +1,382 @@
...
+                predicted,
+                self.greens_function,
+            )
+            # Update the residue
+            residue -= predicted
+            # Add RMS of the residue to the RMSE
+            errors.append(np.sqrt(np.mean(residue ** 2)))
+            # Update source coefficients
+            self.coefs_[point_window] += coeffs_chunk
+        self.errors_ = np.array(errors)
```
leouieda, line 300: How about `rms_per_iteration_` so that it's more informative?

harmonica/equivalent_sources/gradient_boosted.py
leouieda: Makes perfect sense and it says so in the docstring which is perfect. How about just `shuffle` since the function is already `_create_windows`?

harmonica/tests/test_gradient_boosted_eqs.py
```diff
@@ -0,0 +1,263 @@
...
+    # The interpolation should be sufficiently accurate on the data points
+    eqs = EquivalentSourcesGB(window_size=1e3, depth=1e3, damping=None, random_state=42)
+    eqs.fit(coordinates, data)
+    npt.assert_allclose(data, eqs.predict(coordinates), atol=0.02 * vd.maxabs(data))
+
+    # Gridding onto a denser grid should be reasonably accurate when compared
+    # to synthetic values
+    grid = vd.grid_coordinates(region, shape=(60, 60), extra_coords=0)
+    true = point_mass_gravity(grid, points, masses, field="g_z")
+    npt.assert_allclose(true, eqs.predict(grid), atol=0.02 * vd.maxabs(true))
```
leouieda, line 190: With adding a comment that the error tolerance is 2% of the maximum data. And probably best to set `rtol=0` since it defaults to `1e-5`.

leouieda (commented):
This is great! Can't wait to use it 🙂 Left a few suggestions but feel free to reject them if your disagree. This is good to merge 👍

examples/equivalent_sources/gradient_boosted.py
santisoler: Now that you put it in an example, it actually looks better as a regular method. I'll change it!

harmonica/equivalent_sources/gradient_boosted.py
santisoler: Agree! Just a slight modification: `rmse_per_iteration_`, where the `e` stands for "error"

harmonica/tests/test_gradient_boosted_eqs.py
santisoler: > Makes perfect sense and it says so in the docstring which is perfect. How about just shuffle since the function is already _create_windows?

I used `shuffle` at first, but the linter warned me that `shuffle` was taken by the import of `sklearn.utils.shuffle` function. I agree with renaming the argument, so I would need to import the function with an alias.

doc/api/index.rst
```diff
@@ -25,6 +25,7 @@ Equivalent Sources
 
     EquivalentSources
     EquivalentSourcesSph
+    EquivalentSourcesGB
```
santisoler, line 28: Change order of EquivalentSourcesGB in doc/api/index.rst
```suggestion
    EquivalentSourcesGB
    EquivalentSourcesSph
```

harmonica/equivalent_sources/gradient_boosted.py
leouieda: Better than that would be import the module instead `from … import`. That’s always preferred to importing single functions since it keeps the namespaces separate.

examples/equivalent_sources/gradient_boosted.py
```diff
@@ -0,0 +1,132 @@
...
+    block_size=block_size,
+    random_state=42,
+)
+
+# Let's estimate the memory required to store the largest Jacobian when using
+# these values for the window_size and the block_size.
+jacobian_req_memory = eqs_gb.estimate_required_memory(coordinates)
+print(f"Required memory for storing the largest Jacobian: {jacobian_req_memory} bytes")
+
+# Fit the sources coefficients to the observed magnetic anomaly.
```
santisoler, line 81:
```suggestion
# Fit the sources coefficients to the observed gravity anomaly.
```

examples/equivalent_sources/gradient_boosted.py
```diff
@@ -0,0 +1,132 @@
...
+# requires the height of the grid points (upward coordinate). By passing in
+# 1000 m, we're effectively upward-continuing the data.
+grid = eqs_gb.grid(
+    upward=1000,
+    spacing=2e3,
+    data_names="gravity_disturbance",
+)
+print(grid)
+
+# Plot original magnetic anomaly and the gridded and upward-continued version
```
santisoler, line 101:
```suggestion
# Plot original gravity anomaly and the gridded and upward-continued version
```

examples/equivalent_sources/gradient_boosted.py
```diff
@@ -0,0 +1,132 @@
...
+fig, (ax1, ax2) = plt.subplots(nrows=1, ncols=2, figsize=(12, 9), sharey=True)
+
+# Get the maximum absolute value between the original and gridded data so we
+# can use the same color scale for both plots and have 0 centered at the white
+# color.
+maxabs = vd.maxabs(data.gravity_disturbance, grid.gravity_disturbance)
+
+ax1.set_title("Observed magnetic anomaly data")
```
santisoler, line 109:
```suggestion
ax1.set_title("Observed gravity anomaly data")
```

santisoler (commented):
Replace "magnetic" for "gravity" in the example

PR #278, opened 2021-11-10
Add dtype argument to EquivalentSources

santisoler:
The `dtype` argument is used to cast the inputs of the `fit` method, to build
the location of the point sources, to allocate the Jacobian matrix and
therefore to produce predictions. Add a new utility function for casting the
inputs of the fit method to the desired `dtype`. Add new test file for the
equivalent sources utility functions. Add tests for the new feature. Simplify
test suite for equivalent sources through fixtures.

**Reminders**:

### DIFF
--- a/harmonica/tests/test_eq_sources_cartesian.py
+++ b/harmonica/tests/test_eq_sources_cartesian.py
@@ -23,22 +23,6 @@ from ..equivalent_sources.utils import jacobian_numba_serial
 from .utils import run_only_with_numba
 
 
-def build_sample_sources_and_data(region, coords_shape=(40, 40)):
-    """
-    Build a set of sample point sources, data and observation points
-    """
-    # Build synthetic point masses
-    points = vd.grid_coordinates(region=region, shape=(6, 6), extra_coords=-1e3)
-    masses = vd.datasets.CheckerBoard(amplitude=1e13, region=region).predict(points)
-    # Define a set of observation points
-    coordinates = vd.grid_coordinates(region=region, shape=coords_shape, extra_coords=0)
-    # Get synthetic data
-    data = point_mass_gravity(coordinates, points, masses, field="g_z")
-    # Create synthetic weights
-    weights = np.ones_like(data)
-    return points, masses, coordinates, data, weights
-
-
 @pytest.fixture(name="region")
 def fixture_region():
     """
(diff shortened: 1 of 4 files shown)


harmonica/tests/test_eq_sources_cartesian.py
```diff
@@ -407,3 +405,69 @@ def test_backward_eqlharmonic(depth_type):
...
+    )
+    eqs.fit(coordinates, data, weights)
+    # Make some predictions
+    prediction = eqs.predict(coordinates)
+    # Check data type of created objects
+    for coord in eqs.points_:
+        assert coord.dtype == np.dtype(dtype)
+    assert prediction.dtype == np.dtype(dtype)
+    # Check the data type of the source coefficients
+    #  assert eqs.coefs_.dtype == np.dtype(dtype)
```
santisoler, line 449: @leouieda I've noticed that the `coefs_` are casted to `float64` even if the inputs to `vdb.least_squares` are all `float32`. Even that, the predictions are created as `float32`.

leouieda: I was trying to track this down. It seems that scikit-learn uses the dtype of the Jacobian for everything. But the coeffs are actually coming from [scipy.linalg.solve](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.solve.html#scipy.linalg.solve) according to [this line](https://github.com/scikit-learn/scikit-learn/blob/0d378913be6d7e485b792ea36e9268be31ed52d0/sklearn/linear_model/_ridge.py#L156). The docs for `solve` don't mention dtype so it could be that they always return a `float64` from the underlying LAPACK functions. 

It would be fine to leave the coefs the way they are returned since they usually don't require much storage. It's mostly the Jacobian that's affected.

leouieda:
Any thoughts on defaulting to float32?

santisoler:
> Any thoughts on defaulting to float32?

I ran this little script to compare the accuracy of float32 results:
```python
import verde as vd
import harmonica as hm

region = (-3e3, -1e3, 5e3, 7e3)
shape = (40, 40)
coordinates = vd.grid_coordinates(region=region, shape=shape, extra_coords=0)
points = vd.grid_coordinates(region=region, shape=(6, 6), extra_coords=-1e3)
masses = vd.datasets.CheckerBoard(amplitude=1e13, region=region).predict(points)
data = hm.point_mass_gravity(coordinates, points, masses, field="g_z")

eqs64 = hm.EquivalentSources(dtype="float64")
eqs64.fit(coordinates, data)
prediction_64 = eqs64.predict(coordinates)

eqs32 = hm.EquivalentSources(dtype="float32")
eqs32.fit(coordinates, data)
prediction_32 = eqs32.predict(coordinates)


fig, ((ax1, ax2), (ax3, ax4)) = plt.subplots(nrows=2, ncols=2, sharex=True, sharey=True)

maxabs = vd.maxabs(prediction_32, prediction_64)

pc = ax1.pcolormesh(
    *coordinates[:2], prediction_32, vmin=-maxabs, vmax=maxabs, cmap="seismic"
)
plt.colorbar(pc, ax=ax1)
ax1.set_title("Prediction 32bits")
ax1.set_aspect("equal")

pc = ax2.pcolormesh(
    *coordinates[:2], prediction_64, vmin=-maxabs, vmax=maxabs, cmap="seismic"
)
plt.colorbar(pc, ax=ax2)
ax2.set_title("Prediction 64bits")
ax2.set_aspect("equal")

diff = data - prediction_32
maxabs = vd.maxabs(diff)

pc = ax3.pcolormesh(*coordinates[:2], diff, vmin=-maxabs, vmax=maxabs, cmap="seismic")
plt.colorbar(pc, ax=ax3)
ax3.set_title("Diff 32bits")
ax3.set_aspect("equal")

diff = data - prediction_64
maxabs = vd.maxabs(diff)

pc = ax4.pcolormesh(*coordinates[:2], diff, vmin=-maxabs, vmax=maxabs, cmap="seismic")
plt.colorbar(pc, ax=ax4)
ax4.set_title("Diff 64bits")
ax4.set_aspect("equal")

plt.show()
```

With the following result:

float64_vs_float32

The errors of the float32 are 5 orders of magnitude higher than using float64. In the light of these results, I would say to keep float64 by default.

In case the dataset is large enough so it won't fit in memory, but not soooo large, would be interesting to see a comparison between using float32 and gradient-boosted equivalent sources. I suspect float32 would give better results.

leouieda:
👍 sounds good to me! We can add a tutorial about this at some point. Merge away!

PR #282, opened 2021-11-16
Replace conda for pip on Actions and split requirements files

### LINKED ISSUE #257: Refactor CIs by splitting package requirements as needed
Would be nice to follow the latest scheme for running CIs present in Boule, where there's a global `requierements.txt` file with all needed dependencies by the pacakge, while there's an additional `env` directory that hosts several `requirementes_*.txt` files, each one with the dependencies for running only particular stuff: running tests, bulding docs, etc.

The GitHub Actions should be then updated to use these new files.
Would be nice to run tests only using `pip`, since we don't need any fancy package in Harmonica.
But for building the docs we do need `cartopy` and a few other dependencies that are much easily installed through `conda`.

santisoler:
Create a new `env` directory with some `requirements-*.txt` files, each one
with the specific dependencies to run certain tasks, like checking style,
running the test, publishing to PyPI and building the docs. The GitHub Actions
now use `pip` to install the dependencies instead of having to setup Miniconda
and install them through `conda`. The only exception is the one that builds the
docs: we need `pyproj` and `cartopy` to do so and they are easier to install
through `conda`.

Fixes #257 

**Reminders**:


.github/workflows/test.yml
```diff
@@ -78,56 +74,48 @@ jobs:
...
-          auto-update-conda: true
-          channels: conda-forge
-          show-channel-urls: true
-          activate-environment: testing
-          # Needed for caching
-          use-only-tar-bz2: true
+          path: ${{ steps.pip-cache.outputs.dir }}
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements*.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
```
leouieda, line 93:
```suggestion
            ${{ runner.os }}-pip-${{ hashFiles('**/requirements*.txt') }}
```

Seems reasonable right? That way we restore the cache if the requirements haven't changed?

leouieda (approved):
Looks good to me! Left a comment on the pip cache part but I'm not sure if it's relevant.

.github/workflows/test.yml
santisoler: Yes! Thanks for the catch!

PR #288, opened 2021-11-24
Add computation of gravitational tensor components for point sources

santisoler:
Add new kernel functions to compute gravity tensor components generated by
point sources. Add test functions for the new feature: check that the diagonal
elements satisfy the Laplace equation, compare all components against finite
difference computations from the gravity acceleration. Add test class for
checking the symmetry of tensor components. Refactor old test functions for
point gravity: merge some functions into single ones through pytest
parametrizations. Avoid using "gradient" for specifying the gravity
acceleration vector: the "gravity gradient" is usually used to refer to the
tensor.

**Reminders**:


santisoler:
@leouieda I finally found a very good opportunity to add a test class! The tests for tensor components symmetry has a few different cases that were hard to cover with just a few pytest parameters. So I decided it would be better to define a class for it, with different methods for each one of these cases. Let me know what do you think.

harmonica/tests/test_point_gravity.py
```diff
@@ -176,252 +176,438 @@ def test_potential_cartesian_symmetry():
...
+    # Convert to mGal
+    finite_diff *= 1e5
     # Remember that the ``g_z`` field returns the downward component of the
     # gravitational acceleration. As a consequence, the numerical
-    # derivativative is multiplied by -1.
-    approximated_deriv = -1e5 * (potential[1] - potential[0]) / (2.0 * delta)
-
-    # Compute the relative error
-    relative_error = np.abs((approximated_deriv - exact_deriv) / exact_deriv)
+    # derivative dive is multiplied by -1.
```
leouieda, line 257:
```suggestion
    # derivative is multiplied by -1.
```

leouieda (approved):
Looks good to me! I had never used a class for this before. Is the advantage only being able to hold values as attributes? Not complaining, just curious 🙂 

Caught a little typo but otherwise merge away!

santisoler:
> I had never used a class for this before. Is the advantage only being able to hold values as attributes? Not complaining, just curious

Neither do I. I saw that xarray uses classes in their tests, mostly to group and keep tests organized and use inheritance to avoid repeating themselves. I thought that here this new class would work in that sense: group all tensor symmetry related fixtures, parameters and functions (methods now) under the same structure. I think the current scheme improves readability. Pytest parametrizations are nice to avoid repetition and buggy `for` loops inside test functions, but for some multi parameters cases, they fell short and create harder to read code.

To finish this PR I would:
- [ ] Add an example (or wait for the improvement of the docs and add it as a part of the User Guide)
- [ ] Check if passing `g_ne` is the same as passing `g_en`. If not, maybe add a utils function to sanitize tensor component names (we will probably use it in other forward modelling functions).

PR #291, opened 2021-12-01
Add conversion of prisms or a prism layer to PyVista objects

santisoler:
Add a new `visualization` module that hosts `prism_to_pyvista`: a function to
convert a set of prisms into a `pyvista.UnstructuredGrid`. Include the new
module and this function in the API Reference. Add a new `to_pyvista()` method
to the `PrismLayer` accessor that converts a prism layer into a `pyvista` grid,
making it easier to plot it in 3D. The `UnstructuredGrid` has the information
about each prism as hexahedrons, along with their physical properties as cell
data. Add tests for the new features. Add `pyvista` and `vtk` as optional
dependencies to `environment.yml` and `setup.cfg`. Add a new example for
plotting a `PrismLayer`. Configure Sphinx to show pyvista plots in the gallery
and to use the `pyvista-plot` directive in docstrings.

**Reminders**:


leouieda:
@santisoler how about posting this to the PyVista forum to get feedback from the devs? https://github.com/pyvista/pyvista/discussions

santisoler:
Sure! I would love to have feedback from them.

examples/visualization/prism_layer.py
```diff
@@ -0,0 +1,87 @@
...
+prisms = hm.prism_layer(
+    (south_africa_topo.easting, south_africa_topo.northing),
+    surface=south_africa_topo,
+    reference=0,
+    properties={"density": density},
+)
+
+# Create a pyvista UnstructuredGrid from the prism layer
+pv_grid = prisms.prism_layer.to_pyvista()
+print(pv_grid)
```
banesullivan, line 63: Perhaps use Sphinx-Gallery's repr capturing here instead of printing: https://sphinx-gallery.github.io/stable/auto_examples/plot_3_capture_repr.html#capturing-output-representations

```suggestion
pv_grid

###############################################################################
```

harmonica/visualization/prism.py
```diff
@@ -0,0 +1,110 @@
...
+
+    Returns
+    -------
+    pv_grid : :class:`pyvista.UnstructuredGrid`
+        :class:`pyvista.UnstructuredGrid` that represents the prisms with their
+        properties (if any).
+    """
+    # Check if pyvista are installed
+    if pyvista is None:
+        raise ValueError(
```
banesullivan, line 45: This should be an `ImportError`

```suggestion
        raise ImportError(
```

harmonica/visualization/prism.py
```diff
@@ -0,0 +1,110 @@
...
+            "Missing optional dependency 'pyvista' required for building pyvista grids."
+        )
+    # Get prisms and number of prisms
+    prisms = np.atleast_2d(prisms)
+    n_prisms = prisms.shape[0]
+    # Get the vertices of the prisms
+    vertices = np.vstack([_get_prism_vertices(prism) for prism in prisms])
```
banesullivan, line 52: Hm... I wonder if there is a cost to this list comprehension for large meshes. Perhaps there is a way to vectorize this with NumPy?

harmonica/visualization/prism.py
```diff
@@ -0,0 +1,110 @@
...
+    # Generate the cells for the hexahedrons
+    cells = np.arange(n_prisms * 8).reshape([n_prisms, 8])
+    # Build the UnstructuredGrid
+    pv_grid = pyvista.UnstructuredGrid({vtk.VTK_HEXAHEDRON: cells}, vertices)
+    # Add properties to the grid
+    if properties is not None:
+        for name, prop in properties.items():
+            # Assign the property to the cell_data
+            pv_grid.cell_data[name] = np.atleast_1d(prop).ravel()
```
banesullivan, line 61: Just a drive by note: you may want to double check that the array ordering is consistent here. Most of the time folks want `.ravel(order='F')` instead of the default `C` ordering. However since you are generating the cells for the UGrid explicitly, everything likely lines up

banesullivan (commented):
This is awesome work!! And it really goes to show just how simply PyVista can make generating complex UnstructuredGrids over raw VTK

doc/conf.py
```diff
@@ -94,9 +96,22 @@
     "doc_module": "harmonica",
     # Insert links to documentation of objects in the examples
     "reference_url": {"harmonica": None},
+    # Add pyvista to the image scrapers
+    "image_scrapers": ("pyvista"),
```
banesullivan, line 100: This would actually remove the default MPL scraper. Be sure to include it as well

```suggestion
    "image_scrapers": ("pyvista", "matplotlib"),
```

examples/visualization/prism_layer.py
santisoler: Nice! Thanks for the suggestion.

harmonica/visualization/prism.py
santisoler: I'm sure there is place for optimization on this step. We would need to allocate this new array since our prisms are given as 6 boundaries and the vertices need the 3 coordinates of each one of the 8 vertices. But might be a way to create this array in a vectorized way without list comprehensions.

Let me experiment with it and give it a try.

harmonica/visualization/prism.py
santisoler: I like how you are bringing a different point of view into it. I decided to use the same approach we are using on the prism forward modelling (if the density array is 2d, we ravel it), so both functions could take the same inputs and produce objects with the same logic.
https://github.com/fatiando/harmonica/blob/c1c2add5b0a40fd0104d5859b2d0048696f1954a/harmonica/forward/prism.py#L124

But you're right we are making an assumption here, and if any user is assuming the opposite, they wouldn't get a hint that something might be wrong with their results.
Maybe this is something we need to change project-wide. For now I think I can remove the raveling and ask for 1d properties arrays only. The `to_pyvista` method for the prism layer accessor would need to ravel it by itself, but in there we already have information about the correspondence between prism and its properties.

@leouieda  what do you think?

doc/conf.py
santisoler: Thanks for that! I was just copy-pasting your docs.

santisoler:
Thanks a lot @banesullivan for the review! I really appreciate it!

I'm stepping ahead of time, but after this PR is merged, I want to start drafting a similar way to plot tesseroids with pyvista. That might be more challenging since two of their faces are actually spherical. I'll surely ping you when I have some draft :eyes:

harmonica/visualization/prism.py
leouieda: I think this function should require 1D arrays since the prisms already have to be an unstructured 1D "list" of prisms anyway. Let the user ravel if needed.

leouieda:
@santisoler I'm wondering if a new subpackage is really needed here. We're not planning on adding a bunch of visualizations to this package and this is a bit of an exception since we couldn't avoid it. Would it be better to leave that code in the `prism_layer.py` module for now? Or at least this subpackage should be private (`_visualization`) since we don't want people importing it directly.

harmonica/visualization/prism.py
banesullivan: FWIW, +1 for leaving the reveling to users - always best to let users have control and be able to easily choose F vs C

santisoler:
I'm retaking this PR after a **long** time!

> @santisoler I'm wondering if a new subpackage is really needed here. We're not planning on adding a bunch of visualizations to this package and this is a bit of an exception since we couldn't avoid it. Would it be better to leave that code in the prism_layer.py module for now? Or at least this subpackage should be private (_visualization) since we don't want people importing it directly.

The way the `prisms_to_pyvista` function is written, it would allow users to pass any set of prisms and create an UnstructuredGrid that represent them.
To do so, we should allow them to use this function as `hm.prisms_to_pyvista` or (not a fan) `hm.visualization.prisms_to_pyvista`.

Also, as mentioned before, I'm planning to add visualization functions for tesseroids as well (converstion to PyVista objects), so the `visualization` module sounds a good place for them as well.

If we decide to only allow these conversions through the methods of `PrismLayer` (and the future `TesseroidLayer`), so I'm ok with making `_visualization` private. But since it's already defined for the general purpose of any random set of prisms, I think we should keep it public. What do you think?

harmonica/visualization/prism.py
santisoler: Thanks for the feedback. Done! 👍🏼

harmonica/visualization/prism.py
santisoler: I've just pushed a more efficient way to generate the vertices using Numpy. I only iterate with Python for loops along the 6 boundaries (west, east, south, north, bottom, top) and on the 4 vertices that each one of the appear, leading to only 24 inefficient iterations, regardless of the size of the set of prisms.

harmonica/visualization/prism.py
```diff
@@ -0,0 +1,215 @@
...
+                    + "Please, pass prism properties as 1d arrays."
+                )
+            # Assign the property to the cell_data
+            pv_grid.cell_data[name] = prop
+    return pv_grid
+
+
+def _prisms_boundaries_to_vertices(prisms):
+    """
+    Converts prisms boundaries to set of vertices for each prism
```
santisoler, line 104:
```suggestion
    Converts prisms boundaries to sets of vertices for each prism
```

santisoler:
I've made the `visualization` module public after what we decided on the [Fatiando Call of 2022-04-22](https://github.com/fatiando/community/blob/main/development-calls/2022.md#2022-04-22).

santisoler:
Merged even codecov was failing. We are missing the test in cases where `pyvista` is missing and cannot be imported. We can safely ignore that.

PR #295, opened 2021-12-20
Pin style checkers and formatters

leouieda:
Leaving them unpinned, particularly when we have so many now is going to
be a nightmare. Better to have them pinned in the environment.yml and
env/requirements-style.txt files and we upgrade them from time to time.
Fixes a false-positive from the updated RST linter.

Could be a good excuse for us to finally setup dependabot to send PRs when new versions of the linter come out.

**Reminders**:


santisoler:
Thanks for pinning them! I'm merging this.

PR #296, opened 2021-12-20
Move configuration from setup.py to setup.cfg

leouieda:
Make the move away from setup.py following the recommendations from the
Python packaging guides. Moves the requirement listing to setup.cfg as
well and will use a script to extract this for conda installing on CI.

**Reminders**:


leouieda:
Linting error is fixed in #295

leouieda:
Done!

Should I update the style checking workflow to match this? https://github.com/fatiando/pooch/blob/main/.github/workflows/style.yml  

Basically concentrate on a "format" and a "style" run, that way if format fails we know that the solution is "run make format" (no matter why it failed).

Otherwise, this is only held back by the linting false-positive.

leouieda:
Also applies https://github.com/fatiando/community/issues/42

santisoler (approved):
Looking great! After long time, I'm merging this.

PR #299, opened 2022-01-17
Add more FFT-based transformations and filters

LL-Geo:
Add FFT-based transformations and filters for horizontal derivatives, upward continuation, reduction to the pole of magnetic grids, and low-pass and high-pass Gaussian filters. Add test functions for each filter and transformation. Add new functions to the API reference.

**Reminders**:


LL-Geo:
There is an example of how this code works.
https://nbviewer.org/github/LL-Geo/PFToolbox/blob/master/FFT_Filter.ipynb

LL-Geo:
I'll work on the documentation (formula) & formate later... 
Any suggestions are welcome 😄

RichardScottOZ:
Thanks, will see what I can do later today.

RichardScottOZ:
Sent you a pull request with a few small updates.

RichardScottOZ:
@santisoler work here https://github.com/fatiando/harmonica/blob/fft-derivatives/harmonica/filters/fft.py for reference

LL-Geo:
Opps.. commit to wrong branch 🙈

LL-Geo:
> > Opps.. commit to wrong branch see_no_evil
> 
> Ha! No big deal, this can happen to everyone! But since `git` is so good, you can easily solve this with `cherry-pick`:
> 
> 1. Switch to your `isostasy` branch.
> 2. Run `git cherry-pick 3e4d14104c7db3c32fb4a3a95a39e935bf1f9d3a`: git will find that commit across your branches and try to include it to the current branch.
> 3. Check if the commit has been correctly added to `isostasy` with `git log`.
> 4. Push `isostasy`
> 
> And for this branch you can use `revert`:
> 
> 1. Switch to your `main` branch.
> 2. Run `git revert 3e4d14104c7db3c32fb4a3a95a39e935bf1f9d3a`
> 3. And then push

Thanks @santisoler ! I finally remember to do this.. 😅

LL-Geo:
I have added all filters in the new format. @santisoler Just curious how do we do the test? Do we want to do it by comparing the FFT result with the forwarding result, or do we just test the filter itself?  😆 
One silly question about the code style. Do we do `from_ .filters._filters import *` do `from .filters._filters import derivative_upward, derivative_easting`. The list will be very long...

LL-Geo:
We might need to make the magnetic point source work first then we could test RTP and pseudo_gravity... And how could we test the Gaussian filter?..

RichardScottOZ:
I suppose a practical usage consideration, too - memory use compare to size of dataset?

e.g. if say pseudogravity is doing more than one thing, what will it use?

RichardScottOZ:
@LL-Geo I noticed on the Pseudogravity

image

RichardScottOZ:
transform is wrong in Y?

santisoler:
> transform is wrong in Y?

@RichardScottOZ I think there's a missing `return da_filter` in the `pseudo_gravity_kernel` function:
https://github.com/LL-Geo/harmonica/blob/5416778c88a5591030c765c203aa5adc54460c7b/harmonica/filters/_filters.py#L437-L528

Let me add it and give it another try.

BTW, @RichardScottOZ, are you using this branch on production already? If so, I'll try to prioritize this PR.

santisoler:
> We might need to make the magnetic point source work first then we could test RTP and pseudo_gravity... 

Got it. I'll be working on the kernels for the magnetic dipoles in Choclo. When those are up, we can used them to test these filters right away.

> And how could we test the Gaussian filter?

That's trickier. I think if we test the kernel function well enough, we could be confident that the public function will apply it in the right way, since it will be using the `apply_filter` function we are already extensively testing. I wouldn't put a lot of thought on that, at least for now.

RichardScottOZ:
> > transform is wrong in Y?
> 
> @RichardScottOZ I think there's a missing `return da_filter` in the `pseudo_gravity_kernel` function: https://github.com/LL-Geo/harmonica/blob/5416778c88a5591030c765c203aa5adc54460c7b/harmonica/filters/_filters.py#L437-L528
> 
> Let me add it and give it another try.
> 
> BTW, @RichardScottOZ, are you using this branch on production already? If so, I'll try to prioritize this PR.

Yes, been using it quite a lot - but just noticed the wrong way around a week or so ago when I fed a layer into something else - which also needs I have a 50GB pseudograv map of Australia to fix too. :)

RichardScottOZ:
Ever since LL did the sample notebook I have been tinkering.

Let me know when you want me to test something.

LL-Geo:
> > > transform is wrong in Y?
> > 
> > 
> > @RichardScottOZ I think there's a missing `return da_filter` in the `pseudo_gravity_kernel` function: https://github.com/LL-Geo/harmonica/blob/5416778c88a5591030c765c203aa5adc54460c7b/harmonica/filters/_filters.py#L437-L528
> > Let me add it and give it another try.
> > BTW, @RichardScottOZ, are you using this branch on production already? If so, I'll try to prioritize this PR.
> 
> Yes, been using it quite a lot - but just noticed the wrong way around a week or so ago when I fed a layer into something else - which also needs I have a 50GB pseudograv map of Australia to fix too. :)

Oops... Sorry about that...

Have a short break this month. I'll make some updates when I back to office tomorrow. 😄

RichardScottOZ:
Thanks!

RichardScottOZ:
Thanks!

so what is the difference between https://github.com/LL-Geo/PFToolbox/blob/master/FFT_Filter.ipynb
and the changes here?

Don't think I see it at first or second look.

LL-Geo:
> Thanks!
> 
> so what is the difference between https://github.com/LL-Geo/PFToolbox/blob/master/FFT_Filter.ipynb and the changes here?
> 
> Don't think I see it at first or second look.

Yup. These are basically the same.

My test case didn't see the issue for the Y. Could you send me some data for this?

Capture

RichardScottOZ:
Ok, will see what I can find.

RichardScottOZ:
Assuming of course I did it correctly, too.

RichardScottOZ:
image
[gAerogmagnetico_Reducido_al_polo.zip](https://github.com/fatiando/harmonica/files/9662238/gAerogmagnetico_Reducido_al_polo.zip)

RichardScottOZ:
Admittedly just a default test as don't have access to the raw data details

gridpseudo = FFT_Filter(mag.squeeze()).pseudo_gravity(90,0,Im=None,Dm=None,F=50000,savefilter=False)

LL-Geo:
> image [gAerogmagnetico_Reducido_al_polo.zip](https://github.com/fatiando/harmonica/files/9662238/gAerogmagnetico_Reducido_al_polo.zip)

Yup, I see the problem now.

I think the issue come from xrft, which will return an increased order of y coordinate when doing ifft.

image
image

Find similar flip latitude in xrft example https://xrft.readthedocs.io/en/latest/DFT-iDFT_example.html

LL-Geo:
Maybe add a warning use ascending order of y?

The result is actually not wrong (The Y and matrix are all flipped). But need to be careful that sometimes the y coordinate is changed slightly (tiny change). 

image
image

RichardScottOZ:
Yeah - not wrong in that look at the output sense - but wrong if going to do downstream modelling that has affine code type work in it.  That is how I noticed it originally when something that generally worked after going through a bunch of checks didn't, so finally checked that last. Do you think would could add a kwarg that flipped it back if changed?  e.g. keep original y sign as such?

RichardScottOZ:
Something like reset the transform to original and

```python
gridpseudo = gridpseudo.isel(y=slice(None, None, -1))
```

or something along those lines?

RichardScottOZ:
which means anything made like that will be flipped, vertical derivatives, upward continuation, etc?

LL-Geo:
> which means anything made like that will be flipped, vertical derivatives, upward continuation, etc?

Yup~ Every filter result will change the y order.

I made some updates from my notebook that can filpy. https://github.com/LL-Geo/PFToolbox/blob/master/FFT_Filter.ipynb

grid_upward_fft_flipy=FFT_Filter(grid.magnetic_anomaly,[20,20],flipy=**True**).upward_continuation(-10000)

But I guess for the Harmonica package we usually have ascending order for Northing for input? I will leave this PR as it is atm... Can update in the future if needed.

RichardScottOZ:
Thanks, will give it a try shortly.

leouieda:
Hi all, what's the status of this? I've been needing the upward continuation in Harmonica for a project and would be keen to help make it happen.

LL-Geo:
> Hi all, what's the status of this? I've been needing the upward continuation in Harmonica for a project and would be keen to help make it happen.

These functions are ready to go! Just need some tests to merge (Guassan High pass/low pass. RTP and pseudo gravity). I'm thinking since we have the ability to read grd from oasis montaj. We can directly compare filter results.

harmonica/filters/_filters.py
```diff
@@ -60,5 +60,478 @@ def derivative_upward_kernel(fft_grid, order=1):
...
+
+    .. math::
+
+        g(\mathbf{k}) = e^\frac{-|\mathbf{k}| ^ 2}{\mathbf{kw}^2}
+
+    where :math:`\mathbf{k}` is the wavenumber vector
+    (:math:`\mathbf{k} = 2\pi \mathbf{f}` where :math:`\mathbf{f}` is the
+    frequency vector) and `\mathbf{kw}` is the wavenumber of the cutoff
+    wavelength
+    :math:`\mathbf{w}` (:math:`\mathbf{kw} = \frac{2\pi} {\mathbf{w}}`).
```
santisoler, line 242: I think the cutoff wavenumber should be a scalar. And I think we can improve its notation:
```suggestion
        g(\mathbf{k}) = e^\frac{-|\mathbf{k}| ^ 2}{k_\text{cutoff}^2}

    where :math:`\mathbf{k}` is the wavenumber vector
    (:math:`\mathbf{k} = 2\pi \mathbf{f}` where :math:`\mathbf{f}` is the
    frequency vector) and :math:`k_\text{cutoff}` is the cutoff wavenumber:
    :math:`k_\text{cutoff} = \frac{2\pi}{\lambda_\text{cutoff}}`, 
    where :math:`\lambda_\text{cutoff}` is the cutoff wavelength.
```

harmonica/filters/_filters.py
```diff
@@ -60,5 +60,478 @@ def derivative_upward_kernel(fft_grid, order=1):
...
+
+    .. math::
+
+        g(\mathbf{k}) = 1-e^\frac{-|\mathbf{k}| ^ 2}{(2\mathbf{kw})^2}
+
+    where :math:`\mathbf{k}` is the wavenumber vector
+    (:math:`\mathbf{k} = 2\pi \mathbf{f}` where :math:`\mathbf{f}` is the
+    frequency vector) and `\mathbf{kw}` is the wavenumber of the cutoff
+    wavelength
+    :math:`\mathbf{w}` (:math:`\mathbf{kw} = \frac{2\pi} {\mathbf{w}}`).
```
santisoler, line 301: Same here regarding the description of the filter and the definition of the cutoff wavenumber.

harmonica/filters/_filters.py
```diff
@@ -60,5 +60,478 @@ def derivative_upward_kernel(fft_grid, order=1):
...
+    )
+    return da_filter
+
+
+def reduction_to_pole_kernel(fft_grid, i, d, im=None, dm=None):
+    r"""
+    Filter for reduction to the pole in frequency domain
+
+    Return a :class:`xarray.DataArray` with the values of the frequency domain
+    filter for computing the upward derivative. The filter is built upon the
```
santisoler, line 349:
```suggestion
    filter for applying a reduction to the pole on magnetic data. The filter is built upon the
```

harmonica/filters/_filters.py
```diff
@@ -60,5 +60,478 @@ def derivative_upward_kernel(fft_grid, order=1):
...
+    da_filter.data = np.nan_to_num(da_filter.data, posinf=0, nan=0)
+    return da_filter
+
+
+def pseudo_gravity_kernel(fft_grid, i=90, d=0, im=None, dm=None, f=50000):
+    r"""
+    Filter for pseudo gravity in frequency domain
+
+    Return a :class:`xarray.DataArray` with the values of the frequency domain
+    filter for computing the upward derivative. The filter is built upon the
```
santisoler, line 442:
```suggestion
    filter for computing the pseudo gravity. The filter is built upon the
```

harmonica/tests/test_filters.py
```diff
@@ -241,8 +250,161 @@ def test_derivative_upward_kernel(sample_fft_grid, order):
...
+def test_reduction_to_pole_kernel(sample_fft_grid, i=60, d=45, im=45, dm=50):
+    """
+    Check if reduction_to_pole_kernel works as expected
+    """
+    # Transform degree to rad
+    [i, d] = np.deg2rad([i, d])
+
+    if dm is None or im is None:
+        [im, dm] = [i, d]
+    else:
```
santisoler, line 346: These lines aren't needed in the test, since the `im` and `dm` always have values that are not None.

harmonica/tests/test_filters.py
```diff
@@ -241,8 +250,161 @@ def test_derivative_upward_kernel(sample_fft_grid, order):
...
+def test_pseudo_gravity_kernel(sample_fft_grid, i=60, d=45, im=45, dm=50, f=50):
+    """
+    Check if pseudo_gravity_kernel works as expected
+    """
+    # Transform degree to rad
+    [i, d] = np.deg2rad([i, d])
+
+    if dm is None or im is None:
+        [im, dm] = [i, d]
+    else:
```
santisoler, line 383: Same here.

harmonica/tests/test_transformations.py
```diff
@@ -139,3 +235,159 @@ def test_derivative_upward_order2(sample_potential, sample_g_zz):
...
+    second_deriv_ee = xrft.unpad(second_deriv_ee, pad_width)
+    second_deriv_nn = xrft.unpad(second_deriv_nn, pad_width)
+    # Compare g_nn + g_ee against -g_zz (trim the borders to ignore boundary
+    # effects)
+    trim = 6
+    second_deriv_sum = second_deriv_ee + second_deriv_nn
+    second_deriv_sum = second_deriv_sum[trim:-trim, trim:-trim]
+    second_deriv_zz = second_deriv_zz[trim:-trim, trim:-trim]
+    rms = root_mean_square_error(second_deriv_sum, -second_deriv_zz)
+    assert rms < 1e-20
```
santisoler, line 368: I've been using RMS on all the other tests mainly because the outputs of the FFT filters are not exactly the same as the ones computes through forward modelling. So the RMS offers a good compromise of accounting for most of the points, ignoring if some subset of them are actually not that similar (this happens specially in the boundaries).

But for this test I suspect that we don't need to use RMS, maybe we could just use `npt.assert_allclose()` and tweak the `atol` and/or the `rtol` if needed. I usually plot both to set a floor of error (`atol`) and try to keep the relative error to the lowest possible.

harmonica/tests/test_transformations.py
```diff
@@ -139,3 +235,159 @@ def test_derivative_upward_order2(sample_potential, sample_g_zz):
...
+    )
+    # Calculate upward continuation and unpad it
+    continuation = upward_continuation(gravity_padded, -10e3)
+    continuation = xrft.unpad(continuation, pad_width)
+    # Compare against g_z_upward (trim the borders to ignore boundary effects)
+    trim = 6
+    continuation = continuation[trim:-trim, trim:-trim]
+    g_z_upward = sample_g_z_upward[trim:-trim, trim:-trim]
+    rms = root_mean_square_error(continuation, g_z_upward)
+    assert rms / np.abs(g_z_upward).max() < 0.015
```
santisoler, line 393: Since the upward continuation is a low-pass filter, maybe we could also test this with `npt.assert_allclose()`. If not it's ok to leave it with the rms.

harmonica/transformations.py
```diff
@@ -44,3 +53,276 @@ def derivative_upward(grid, order=1):
...
+    Parameters
+    ----------
+    grid : :class:`xarray.DataArray`
+        A two dimensional :class:`xarray.DataArray` whose coordinates are
+        evenly spaced (regular grid). Its dimensions should be in the following
+        order: *northing*, *easting*. Its coordinates should be defined in the
+        same units.
+    height : float
+        The height of upward continuation. Value should be negative. Its units
+        are the same units of the ``grid``.
```
santisoler, line 145: I didn't mention it in the filter, but the units of the heights should be the same as the coordinates in the grid and not the grid itself.

harmonica/transformations.py
```diff
@@ -44,3 +53,276 @@ def derivative_upward(grid, order=1):
...
+    d : float in degrees
+        The declination inducing Geomagnetic field.
+    im : float in degrees
+        The inclination of the total magnetization of the anomaly source.
+        Default is i, neglecting remanent magnetization and
+        self demagnetization.
+    dm : float in degrees
+        The declination of the total magnetization of the anomaly source.
+        Default is d, neglecting remanent magnetization and
+        self demagnetization.
```
santisoler, line 259: I'm inclined to rename this parameters to a more readable alternative like `inclination`, `declination`, `source_inclination`, `source_declination`. The same goes for the parameters of its corresponding filter.

santisoler (changes requested):
Hi @LL-Geo! Thanks a lot for the changes you pushed into this PR! We want to merge it as soon as possible so these commits come in great time!

I left a few suggestions, mainly to docstrings and styling. Please take a look at them and let me know what you think.

Regarding the grd files you uploaded, I assume you run some of the filters with Oasis Montaj and uploaded them, right? If that's the case they might serve for testing purposes, although I'm inclined to store them as netCDF files so we don't have to rely on the `load_oasis_grd()` function to run these tests. You can totally use it to create those netCDF files though.

Thanks again for your hard work! This is awesome!

harmonica/filters/_filters.py
```diff
@@ -60,5 +60,478 @@ def derivative_upward_kernel(fft_grid, order=1):
...
+    filter for computing the upward continuation. The filter is built upon the
+    frequency coordinates of the passed ``fft_grid`` and is defined as follows:
+
+    .. math::
+
+        g(\mathbf{k}) = e^{|\mathbf{k}| ^ h}
+
+    where :math:`\mathbf{k}` is the wavenumber vector
+    (:math:`\mathbf{k} = 2\pi \mathbf{f}` where :math:`\mathbf{f}` is the
+    frequency vector) and :math:`h` is the height of the upward continuation.
```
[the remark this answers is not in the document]
LL-Geo: Yup~ Good suggestion!

harmonica/filters/_filters.py
LL-Geo: Yup~

harmonica/tests/test_filters.py
LL-Geo: yup

harmonica/tests/test_filters.py
LL-Geo: yup

harmonica/transformations.py
LL-Geo: I rename them as `inclination`, `declination`, `magnetization_inclination`, `magnetization_declination`. I do feel it is a bit long, but it's okay..

harmonica/transformations.py
LL-Geo: Yup~

harmonica/tests/test_transformations.py
LL-Geo: Yup~

harmonica/tests/test_transformations.py
LL-Geo: since the data object is xarray, I change rms to `xarray.testing.assert_allclose()`

LL-Geo:
Thanks, @santisoler for the detailed review and great suggestions! 👍 

I think it's ready to merge. I do see the black check error, but my local machine could pass this test...

I guess I'll leave the example in another PR and use the new dataset in **Ensaio** 😃 

> Run make black-check
> black --check setup.py harmonica examples data/examples doc/conf.py tools
>would reformat harmonica/filters/_filters.py
>would reformat harmonica/tests/test_filters.py
>Oh no! 💥 💔 💥
>2 files would be reformatted, 67 files would be left unchanged.
>make: *** [Makefile:44: black-check] Error 1
>Error: Process completed with exit code 2.

harmonica/tests/test_transformations.py
santisoler: Much better 👍🏼 !

santisoler:
Thanks @LL-Geo for applying all those changes! I just made a few more:

- Improved the math of the horizontal derivative and the bandpass filters
- Added the new functions to the API reference
- Run `make format`. It seems that you are running and older version of Black (< 22 probably) that doesn't complain about the spacings around power operators in case of simple operands. The latest black requires no space in such case.

I like the way this PR is going! I think there are a couple of things that could be improved though.

The math for the reduction of the pole and the pseudo gravity are quite complex as they are. There are a few repeated cosine and sine calculations that could be simplified, but also having a very long equation makes it hard to review and debug. Have you considered implementing something similar to what we had in the old `fatiando` package: https://legacy.fatiando.org/api/gravmag.transform.html#fatiando.gravmag.transform.reduce_to_pole? Is there any reason why your implementation would be preferred?
Also, these two functions share some computations. Maybe we can avoid repeating ourselves.

Once we settle on that, I would like to add some math to the transformation functions. They should be simple, just highlighting that we are applying the FFT to the given array, applying the filter and then going back with an iFFT. Where possible I would add an explicit version of the filter (bandpass filters, derivatives, etc), but I would avoid it for the RTP and the pseudo gravity.

The third topic that concerns me are the tests. Some of them reimplement the same functions exactly as they are in the public functions. This would produce a 100% test coverage but it's not testing anything really: they would never fail because we're testing two identical snippets of code. For the reduction of the pole I would go for the Choclo route: compute the magnetic field of a dipole on a regular grid with given inclinations and declinations (both for the geomagnetic field and the magnetic moment) and with the ones corresponding to the pole; and then I would run the RTP comparing its output with the one obtained from the forward modelling.
I would like to hear @leouieda's input on this as well.

Some of these tests are tricky, I know. So don't feel bad about them, testing is hard!

LL-Geo:
Finally, get some time to do some updates: 😅 
1.  Simplify the math of pseudo gravity
2.  Update some test filter kernels: For RTP I compare with the old fatiando package; for the rest kernel, I compare with come pre-computed value.
3. With test filter kernel from pre-computed value and test transformation with OM output. I think the test coverage should be okay...

harmonica/filters/_filters.py
```diff
@@ -62,3 +62,598 @@ def derivative_upward_kernel(fft_grid, order=1):
...
+    da_filter = _get_rtp_filter(
+        k_easting,
+        k_northing,
+        inclination,
+        declination,
+        magnetization_inclination,
+        magnetization_declination,
+    )
+    # Deal with inf and nan value
+    da_filter.data = np.nan_to_num(da_filter.data, posinf=0, nan=0)
```
santisoler, line 472: Why do we need this line? Where can we find `nan`s in the filter? And why should we set them to 0 (meaning that we turning that frequency off in the output)? Is it possible to catch any potential singularity before an not recurring to overwriting the `nan`s?

harmonica/filters/_filters.py
```diff
@@ -62,3 +62,598 @@ def derivative_upward_kernel(fft_grid, order=1):
...
+                inclination,
+                declination,
+                magnetization_inclination,
+                magnetization_inclination,
+            )
+            * np.sqrt(k_easting**2 + k_northing**2) ** -1
+        )
+    # Deal with inf and nan value
+    da_filter.data = np.nan_to_num(da_filter.data, posinf=0, nan=0)
+    return da_filter / 149.8 / f
```
santisoler, line 576: The `149.8` sounds a little bit magical. Where does it come from? Is it the factor between the density and the magnitude of the magnetization? If that's the case we should include it as a parameter that could be changed. If it involves a universal constant, would be better to add it to the `harmonica.constants` module instead of hardcoding it here in the function.

harmonica/filters/_filters.py
```diff
@@ -62,3 +62,598 @@ def derivative_upward_kernel(fft_grid, order=1):
...
+        Default is i, neglecting remanent magnetization and
+        self demagnetization.
+    magnetization_declination : float in degrees
+        The declination of the total magnetization of the anomaly source.
+        Default is d, neglecting remanent magnetization and
+        self demagnetization.
+    f : float or 2d-array
+        Ambient field in the study area. It can use the mean ambinent field
+        value in the study area or the real ambient field value in all
+        locations. Default is 50,000 nT.
```
santisoler, line 527: I don't quite like the `f` name for this parameter. Could we rename it?

santisoler (changes requested):
Thanks @LL-Geo for the effort on this PR! It's looking much better after each iteration. I refactored the reduction to the pole and the pseudogravity filters a little bit so we avoid repeating ourselves and to simplify the implementation of the filter: it does the same operations as the one you coded, but it just defines some intermediate variables that come handy.

I left a few comments below, mostly small things. Regarding the tests, there are probably more tests we could add (like forward modelling dipoles, etc), but I don't want to block this PR just because of that. So I think after we solve some little issues here we can merge this and continue working on it once it's on `main`.

Also, having @leouieda's input would be nice 🙂

santisoler:
@LL-Geo I've justnoticed that we have a few lines that are being untested:
- If we pass `magnetization_inclination=None` and `magnetization_declination=None` in the pseudogravity kernel.
- If the inclination is 90 degrees in the pseudogravity kernel.

Would you like to add tests for those cases?

harmonica/filters/_filters.py
```diff
@@ -62,3 +62,598 @@ def derivative_upward_kernel(fft_grid, order=1):
...
+    # Grab the coordinates of the Fourier transformed grid
+    freq_easting = fft_grid.coords[dims[1]]
+    freq_northing = fft_grid.coords[dims[0]]
+    # Convert frequencies to wavenumbers
+    k_easting = 2 * np.pi * freq_easting
+    k_northing = 2 * np.pi * freq_northing
+    # Check if input is RTP field
+    if inclination == 90:
+        # Calculate vertical integral
+        da_filter = np.sqrt(k_easting**2 + k_northing**2) ** -1
```
santisoler, line 560: What happens if `inclination == -90`? Shouldn't we consider it as a separate case as well?

harmonica/filters/_filters.py
LL-Geo: That's a lazy implication that replaces the 0 wavenumbers with 0. I change it to `da_filter.loc[dict(freq_northing=0, freq_easting=0)] = 0`

harmonica/filters/_filters.py
LL-Geo: Sure.. Renamed to `ambient field`

LL-Geo:
> Thanks @LL-Geo for the effort on this PR! It's looking much better after each iteration. I refactored the reduction to the pole and the pseudogravity filters a little bit so we avoid repeating ourselves and to simplify the implementation of the filter: it does the same operations as the one you coded, but it just defines some intermediate variables that come handy.
> 
> I left a few comments below, mostly small things. Regarding the tests, there are probably more tests we could add (like forward modelling dipoles, etc), but I don't want to block this PR just because of that. So I think after we solve some little issues here we can merge this and continue working on it once it's on `main`.
> 
> Also, having @leouieda's input would be nice 🙂

That's magic @santisoler! The code looks so good each time when you make changes 👍.

Small editions and some tests are added!

harmonica/filters/_filters.py
santisoler: I see. I think explicitly changing the value of the filter for the zero wavenumber vector is more explicit and less prone to errors. It looks much better now!

harmonica/filters/_filters.py
santisoler: Got it! Makes sense then.

harmonica/filters/_filters.py
```diff
@@ -62,3 +64,599 @@ def derivative_upward_kernel(fft_grid, order=1):
...
+            k_easting,
+            k_northing,
+            inclination,
+            declination,
+            magnetization_inclination,
+            magnetization_declination,
+        )
+    # Set 0 wavenumber to 0
+    da_filter.loc[dict(freq_northing=0, freq_easting=0)] = 0
+    return da_filter * GRAVITATIONAL_CONST * 1e8 / ambient_field
```
santisoler, line 579: I think this looks much better than before. Although I'm still unsure where the `1e8` is coming from. I suspect there's some portion of it that might be due to the fact that we need to convert the `ambient_field` from nT to T, but I'm still one order of magnitude short haha.

santisoler (commented):
Thanks @LL-Geo for the hard work on this. I've just pushed some minor improvements to the documentation of rtp and pseudogravity.

I left two comments on pseudogravity. Check them out and let me know what do you think.

harmonica/transformations.py
```diff
@@ -44,3 +53,309 @@ def derivative_upward(grid, order=1):
...
+        locations. Default is 50,000 nT.
+
+    Returns
+    -------
+    pseudo gravity : :class:`xarray.DataArray`
+        A pseudo gravity :class:`xarray.DataArray` of the passed``grid``.
+        Its units are the same units of the ``grid`` multiply units of its
+        coordinates. The vertical integral of the reduction to pole magnetic
+        field is normalised by the ambient field. It reflects pseudo gravity of
+        a geological body based on cgs susceptibility units.
```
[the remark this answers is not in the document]
LL-Geo: Yup~I'll change that. That will work by divide a 4 pi

harmonica/filters/_filters.py
LL-Geo: Haha. It's a bit confusing... 
If we look at the formula 9:
image
Where VI in `nT*m,` g in `mGal,` p in `g*cm-3,` F in `nT,` and G in `m3*kg−1*s−2`
So convert from `mGal` to `m*s-2`, there is a `1e-5,` and p from `g*cm-3` to `kg*m3` with a `1e3`
So VI/g=`1e-5`/`1e3` * `(k*F)/(p*G)`=(`1e-8/G`)*`F*k/p`
Thus `pseudogravity= VI * G * 1e8 / F`

LL-Geo: Is that make sense to you? Haha

santisoler: Thanks for the clarification @LL-Geo. I think I understand where the `1e-8` comes from now. But I still have many doubts on the development of the equations in that paper in comparison with Blakely.

For example, the authors in equation 3 assumes that there's no remanent magnetization in the source (magnetization is parallel to the ambient field), but in our implementation of the filter we do allow passing different angles for the magnetization and for the geomagnetic field.

Also, they explicitly mention that the susceptibility is given in cgs units, so I suspect that there might be another unit conversion missing. Moreover, is there a reason why the filter don't ask values for the density and the susceptibility? Why assuming that the ratio between them is equal to 1?

I think the pseudogravity field requires some discussion, not only in its implementation but also on its design. So here's my idea: what if we remove it from this PR and we can include it in a separate one? This way we don't block all the other filters (that are pretty ready to be merged) just because we have some discussions on a single one of them.

Let me know what do you think.

LL-Geo: Haha. Both ways works!
The Blakely's implementation is confusing as well. If you have the book and look at the page 408.
image
image

That implementation is based on 100 kg/m3 per A/m.
The idea of the new implementation is decoding magnetization into the ambient field and also magnetic susceptibility.
I know that is not ideal, because we have remanent and demagnetization. So I put the option to include different magnetization inclinations and declinations. That could refer to as apparent susceptibility. If we don't have remanent, everything is perfect (apparent susceptibility equal to the real susceptibility). But sometimes, we need to have some assumptions in the real world.

 As for the unit transformation. For cgs unit, μ=1, as for SI unit, we need to add a 4 pi in here. So the 1 susceptibility(cgs)= 4 Pi susceptibility(SI).

santisoler: Oh yes, those FORTRAN implementations are not the best, we aim to do something better than that 😎 .

> The idea of the new implementation is decoding magnetization into the ambient field and also magnetic susceptibility.
I know that is not ideal, because we have remanent and demagnetization. So I put the option to include different magnetization inclinations and declinations.

I agree with this! I don't like that a lot of geophysical applications assume that susceptibility is a good representative of magnetization of the sources on its own: remanence do exist in nature and it's more common that we usually assume. The point is that Salem's paper explicitly assumes no remanence in equation 3, where they consider that the magnetization vector is parallel to the ambient field:

$$ \mathbf{J} = k \mathbf{F} $$

or with a notation I'm more familiar with:

$$ \mathbf{M} = \chi \mathbf{F} $$

So their equations wouldn't be compatible with a pseudogravity that do takes remanence into account, am I right?

> As for the unit transformation. For cgs unit, μ=1, as for SI unit, we need to add a 4 pi in here. So the 1 susceptibility(cgs)= 4 Pi susceptibility(SI).

Nice!

Again, I think the pseudogravity filter deserves some more discussion and I wouldn't like to block this PR just because of it. So I retake my suggestion of removing it from this PR and open a new one just for it.

LL-Geo: Ahaha.. Let me explain in this way.
In the real world... We have remanent, and we can measure it and calculate the Koenigsberger Ratio (Q). So we have: 
M=xF + M(remanent) = xF + Q*xF = (1+Q)*x*F
I will call (1+Q)*x as the apparent susceptibility.
So. If no remanent Q is 0, the apparent susceptibility is the real susceptibility. If Q is not 0, what should we do?
The answer is petrophysics!

If we don't know petrophysics, we don't know magnetization (although we have some numerical way for approximation), and Koenigsberger Ratio. I guess that's out of the scope of this filter. Because this filter is just RTP plus vertical integral. And we try to link some physics in here.

I will add some lines in the doc to show the unit of this filter (g/cm-3 per apparent susceptibility). And if they have petrophysics (with Q), they can calculate the real susceptibility on their own!

Let me know how do you think (if you still think we need more discussion, I'll remove pg and open a new PR)! And thanks so much for this @santisoler 👍 😄

RichardScottOZ:
I remember looking at the Blakely pseudograv and being happing that @LL-Geo  was doing it!

harmonica/filters/_filters.py
santisoler: Thanks for explanation @LL-Geo!

I see your point. Although I have to say that the remanence that can be considered when Q is non-zero is remanent magnetization that is parallel to the geomagnetic field. In the general case this doesn't happen: rocks can present magnetization on other directions than $\mathbf{F}$, so we cannot express the magnetization vector $\mathbf{M}$ as a scalar times $\mathbf{F}$.

I would love to continue this discussion on the pseudogravity filter. But at the same time I would like to release the next version of Harmonica this month and I'm thrilled to include the FFT in it.
So I would take your proposal of stripping the PG filter from this PR and open a new one for it.

santisoler:
> I remember looking at the Blakely pseudograv and being happing that @LL-Geo was doing it!

Indeed! @LL-Geo is putting a lot of work into it and he's actually teaching me a lot about it!

harmonica/filters/_filters.py
LL-Geo: Okay dokie! Let's move forward @santisoler !

I have removed the PG filter. It should be ready to go 😄

Once this one is merged, I'll open a new PR for PG to continue our discussion!

.gitignore
```diff
@@ -18,3 +18,22 @@ MANIFEST
...
+harmonica/equivalent_sources/gradient_boosted.py
+harmonica/forward/_tesseroid_utils.py
+harmonica/forward/_tesseroid_variable_density.py
+harmonica/forward/point.py
+harmonica/forward/prism.py
+harmonica/forward/utils.py
+harmonica/tests/test_forward_utils.py
+harmonica/tests/test_point_gravity.py
+harmonica/tests/test_tesseroid.py
+harmonica/tests/test_tesseroid_variable_density.py
```
santisoler, line 39: We really want to keep track of these files 😅 
```suggestion
```

doc/api/index.rst
```diff
@@ -25,7 +25,14 @@ Apply well known transformations regular gridded potential fields data.
...
     :toctree: generated/
 
+    derivative_easting
+    derivative_northing
     derivative_upward
+    upward_continuation
+    gaussian_lowpass
+    gaussian_highpass
+    reduction_to_pole
+    pseudo_gravity
```
santisoler, line 35:
```suggestion
```

doc/api/index.rst
```diff
@@ -35,7 +42,14 @@ Define filters in the frequency domain.
...
     :toctree: generated/
 
+    filters.derivative_easting_kernel
+    filters.derivative_northing_kernel
     filters.derivative_upward_kernel
+    filters.upward_continuation_kernel
+    filters.gaussian_lowpass_kernel
+    filters.gaussian_highpass_kernel
+    filters.reduction_to_pole_kernel
+    filters.pseudo_gravity_kernel
```
santisoler, line 52:
```suggestion
```

harmonica/transformations.py
```diff
@@ -44,3 +52,241 @@ def derivative_upward(grid, order=1):
...
+        demagnetization. Default None.
+    magnetization_declination : float in degrees
+        The declination of the total magnetization of the anomaly source. If
+        None, the ``magnetization_declination`` will be set equal to the
+        ``declination``, neglecting remanent magnetization and self
+        demagnetization. Default None.
+
+    Returns
+    -------
+    RTP : :class:`xarray.DataArray`
```
santisoler, line 273:
```suggestion
    reduced_to_pole_grid : :class:`xarray.DataArray`
```

santisoler (approved):

santisoler:
Awesome job @LL-Geo! I just pushed a few minor fixed, mostly to docstrings. I'm merging this 🚀 🏅 Congrats for the hard work! I know that not only @RichardScottOZ will be happy with these filters!

santisoler:
Thanks @RichardScottOZ for your contributions!

RichardScottOZ:
sure, I have a few things I need to try soon, so excellent

harmonica/filters/_filters.py
RichardScottOZ: is thefe one?

PR #303, opened 2022-04-18
Add warning about future deprecation of datasets module

santisoler:
**Description**

In a future release, Harmonica will ditch the self-hosted sample datasets and
replace them by using Ensaio to fetch them when needed. Add FutureWarnings to
each fetch function in the datasets module and point users to Ensaio. Add test
function for these new FutureWarnings. Add warning in Datasets section in API
Reference.

### DIFF
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -71,6 +71,12 @@ Synthetic models and surveys
 Datasets
 --------
 
+.. warning::
+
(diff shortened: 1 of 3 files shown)


santisoler:
Just noticed that synthetic surveys make use of sample datasets. This deprecation might need some thinking. Closing this PR for now until we give it a better thought.

PR #307, opened 2022-04-30
Updata Airy Isostasy

LL-Geo:
use rock equivalent topography to calculate airy isostasy

**Description**

**Relevant issues/PRs**


santisoler:
Thanks @LL-Geo for this PR! With this it the `isostasy_airy()` function could be used to a broad range of cases.

What I like about this new approach is that we move away from passing absolute heights to focusing on the thicknesses of the layers. While doing so, we don't need to assume the existence of any missing boundary (for example, we currently assume that  ocean water top boundary is always at zero height).

My main concerns are regarding the interface.
First, the `basement_elevation` is a little bit confusing. In the docstring it's described as the topography or bathymetry height. Although its name (`basement_elevation`) doesn't suggest it.
Secondly, I'm quite hesitant to include the information about the layers as two separate arguments. What if we ask for a single `layers` dictionary containing the thickness and density of each layer inside a tuple. For example, if we only have topography above zero:

```python
hm.isostasy_airy(
    layers={
        "topography": (topo_thickness, topo_density),
    },
    basement=0,
)
```

Or, if we have topography, sediments and water:

```python
hm.isostasy_airy(
    layers={
        "topography": (topo_thickness, topo_density),
        "sediments": (sediments_thickness, sediments_density),
        "water": (water_thickness, water_density),
    },
    basement=basement,  # depth to the bottom boundary of the deepest layer
)
```

@leouieda I know you wanted to rethink our Airy isostasy function for a while. What do you think?

LL-Geo:
Thanks @santisoler for the nice comments. 👍 

I make some changes in the docstring to make the basement more clear. I know it is common to use the term topography, but it is mainly because we don't know the sediment thickness, neglect its effect, or with no sediment cover. Basically, basement equals topography if we don't consider sediment. When we consider sedimentary layers, it refers to the top of crystalline basement (topography - sediment thickness). 

For point two, I agree a dictionary is more compact. I change the structure of the interface.

# Case 1 no sediment data_africa case
hm.isostasy_airy(
    basement_elevation=data_africa.topography,
    layers={
                "water": (water_thickness, water_density)
    },
)

#Case 2 with sediment
hm.isostasy_airy(
    basement_elevation=topography-sediment_thickness,
    layers={
                "water": (water_thickness, water_density),
                "sediment": (sediment_thickness, sediment_density)
    },
)
 
In that case, we don't need to pass topography into the layers.

santisoler:
Thanks @LL-Geo for applying those changes! This is looking great! 🥇 

I just pushed a few edits to your branch. I renamed the `basement_elevation` argument to `basement`, made a few tweaks on the docstrings, and refactored a little bit the code of the function.
I've also took the opportunity to refactor some of the old tests we had using nice features of pytest and added a new one that runs on multiple layers.
Let me know what do you think! Feel free to disagree, these edits are just a proposal (we can always revert it and go back to the previous version).

BTW, do you have an .svg version of the figure? We often prefer including .svg files instead of .png because they can be easily edited afterwards.

LL-Geo:
> Thanks @LL-Geo for applying those changes! This is looking great! 🥇
> 
> I just pushed a few edits to your branch. I renamed the `basement_elevation` argument to `basement`, made a few tweaks on the docstrings, and refactored a little bit the code of the function. I've also took the opportunity to refactor some of the old tests we had using nice features of pytest and added a new one that runs on multiple layers. Let me know what do you think! Feel free to disagree, these edits are just a proposal (we can always revert it and go back to the previous version).
> 
> BTW, do you have an .svg version of the figure? We often prefer including .svg files instead of .png because they can be easily edited afterwards.

That looks great @santisoler ! Thanks for all the editing. 👍 
I just change a little bit about the layer density description. as I think the density could also be an array as long as it matches the thickness layer (for example if we consider sediment compaction, some areas will have a higher density). A variable density information may be less meaningful in the data attribution. But I think we can just leave it there.

I'll replace the .PNG to .SVG file tomorrow when I back to my office. There is some path issue for the Github when I'm not using the Uni network. Haha.

santisoler:
Awesome job @LL-Geo! Really like what you did with the extra test function for density arrays, nice usage of pytest fixtures! I made a few tweaks here and there, no deal breaker. I've added the new `isostatic_moho_airy` function to `docs/api/index.rst` so it gets listed on the API Reference. I've also transformed the isostasy subpackage into a single `isostasy.py` file. I think we can host future isostasy functions in that file without any issues. If that's not the case then we can think about creating a subpackage for it.

If the CI gods approves, I think this is ready to be merged. What do you think? Would you like to change something else?

LL-Geo:
> Awesome job @LL-Geo! Really like what you did with the extra test function for density arrays, nice usage of pytest fixtures! I made a few tweaks here and there, no deal breaker. I've added the new `isostatic_moho_airy` function to `docs/api/index.rst` so it gets listed on the API Reference. I've also transformed the isostasy subpackage into a single `isostasy.py` file. I think we can host future isostasy functions in that file without any issues. If that's not the case then we can think about creating a subpackage for it.
> 
> If the CI gods approves, I think this is ready to be merged. What do you think? Would you like to change something else?

That looks great @santisoler 👍 ! Thanks for all the editing and change! Let's merge it, haha.

leouieda:
Really happy to see this merged 🚀

PR #315, opened 2022-05-24
Add progressbar to prism forward gravity calculations

### LINKED ISSUE #312: Adding progress bar to forward gravity calculation of prisms

During a call to `prism.gravity` (or `prism_layer.gravity`) it would be great to have a progress bar which updates as the calculations advance through the gravity observation points (`coordinates`) .  @LL-Geo suggested the `numba_progress` package, which seems to be a good fit. Looking through `harmonica.forward.prisms.py`, I assume implementing this feature would be done within the `jit_prism_gravity` function, and when calling the function, users would use something along the lines of:

``` python
with ProgressBar(total=max_iter) as numba_progress:
        topo_prisms.prism_layer.gravity(
               coordinates=(df_grav.x, df_grav.y, df_grav.z),
               field = 'g_z',
               progress=numba_progress)
```

Yes I'm willing to try, but will likely need some assistance since it's my first code contribution, and I'm unfamiliar with numba_progress.

mdtanker:
**Description**

This PR adds an optional progress bar to the forward gravity calculation of `prisms_gravity`. It uses `numba_progress`, and is called with a flag `progressbar=True`, which defaults to `False`. Currently it only works properly with `parallel=True`.

**Relevant issues/PRs**

Fixes #312


harmonica/forward/prism.py
```diff
@@ -73,6 +74,10 @@ def prism_gravity(
     dtype : data-type (optional)
         Data type assigned to the resulting gravitational field. Default to
         ``np.float64``.
+    progressbar : bool (optional)
+        If True a progress bar for the number of gravity coordinates to solve
+        in the forward calculation will be displayed using the numba-progress
+        package. Default to ``False``.
```
santisoler, line 80: Small changes in the docstrings to make it clearer how the progressbar will be printed.
BTW, we could add a link to `numba_progress` to the `intersphinx_mapping` dict in `doc/conf.py` so the `numba_progress` is clickable and links to its website.

```suggestion
        If True, a progress bar of the computation will be printed to standard error (stderr).
        Requires :mod:`numba_progress` to be installed.
        Default to ``False``.
```

harmonica/forward/prism.py
```diff
@@ -130,9 +135,26 @@ def prism_gravity(
...
+    # Show progress bar for 'jit_prism_gravity' function
+    if progressbar:
+        try:
+            from numba_progress import ProgressBar
+
+            progress_proxy = ProgressBar(total=coordinates[0].size)
+        except ImportError:
+            raise ImportError(
+                "Missing optional dependency 'numba_progress' required if progressbar=True"
+            )
```
santisoler, line 147: This works great, but it could be improved.We don't need to try to import `ProgressBar` on every function call, we can move the import step outside the function.

So, I would take the import outside the function, but keeping the `try`, `except` statements to avoid `ImportError` on an optional dependency:

```python
try:
    from numba_progress import ProgressBar
except ImportError:
    ProgressBar = None
```

And inside the function call, we could just raise an `ImportError` if `progressbar` is set to `True` but `ProgressBar` is `None`:

```python
progress_proxy = None
if progressbar:
    if ProgressBar is None:
        raise ImportError("... <custom message> ...")
    progress_proxy = ProgressBar(total=coordinates[0].size)
```

What do you think? Does it sounds better?

harmonica/forward/prism.py
```diff
@@ -186,7 +208,7 @@ def _check_prisms(prisms):
         raise ValueError(err_msg)
 
 
-def jit_prism_gravity(coordinates, prisms, density, kernel, out):
+def jit_prism_gravity(coordinates, prisms, density, kernel, out, progress_proxy):
```
santisoler, line 211: Is it possible to make `progress_proxy` an optional parameter with default to `None`? We are not currently obliged to keep backwards compatibility, and specially not with private functions, but since it's not required to perform the computation, sounds better if we keep it optional. If Numba complains about it, we can go back, no problem.

```suggestion
def jit_prism_gravity(coordinates, prisms, density, kernel, out, progress_proxy=None):
```

harmonica/forward/prism_layer.py
```diff
@@ -22,6 +22,7 @@ def prism_layer(
     surface,
     reference,
     properties=None,
+    progressbar=False,
```
santisoler, line 25: This argument is unused and we don't actually need it. Users that want a progressbar while computing the gravity fields of a prism layer would need to pass it to the `gravity` method of the Dataset Accessor.

```suggestion
```

harmonica/forward/prism.py
```diff
@@ -212,6 +234,9 @@ def jit_prism_gravity(coordinates, prisms, density, kernel, out):
     """
     # Iterate over computation points and prisms
     for l in prange(coordinates[0].size):
+        # Update progress bar if called
+        if progress_proxy is not None:
+            progress_proxy.update(1)
```
santisoler, line 239: I think it might be better to move these lines to the bottom of the for loop: if for some reason the very last iteration fails, the progressbar will still show 100%. But if we move these lines at the end,reaching the 100% is done only after the last iteration finishes. Do you agree?

harmonica/forward/prism.py
```diff
@@ -212,6 +234,9 @@ def jit_prism_gravity(coordinates, prisms, density, kernel, out):
     """
     # Iterate over computation points and prisms
     for l in prange(coordinates[0].size):
+        # Update progress bar if called
+        if progress_proxy is not None:
```
santisoler, line 238: Thinking about some little optimizations.

I imagine that checking if `progress_proxy` is `None` on each iteration is slower than checking if a boolean is True or False. So, I'm thinking we could do something like:

```python
# Check if we need to update the progressbar on each iteration
update_progressbar = (progress_proxy is not None)
# Iterate over computation points and prisms
    for l in prange(coordinates[0].size):
        for m in range(prisms.shape[0]):
            ...
        # Update progress bar if called
        if update_progressbar:
            progress_proxy.update(1)
```

I'm not sure if it would run faster, but might worth experimenting maybe.
cc @leouieda

santisoler (changes requested):
Looking awesome @mdtanker !

I left some small suggestions to improve the code. Nothing critical, just a few ideas.

Besides those, the only thing we need is to write some tests for this. Feel free to ask for help if you need some!

harmonica/forward/prism.py
mdtanker: Good call, I'll move them down.

harmonica/forward/prism.py
mdtanker: That seems like a better option. I've just done a quick test and it appears to be faster, ~7%, but that was just a single test.

harmonica/forward/prism_layer.py
mdtanker: Okay I thought that was necessary, but it appears to work fine without.

harmonica/forward/prism.py
mdtanker: Yes that seems to still work so I'll add your suggestion in.

harmonica/forward/prism.py
mdtanker: Yeah, that makes sense, thanks. I've added your suggestion.

mdtanker:
Thanks for all the suggestions @santisoler. I've added them all in and have pushed the new commits. Any guidance for the tests would be appreciated!

harmonica/forward/prism.py
santisoler: Awesome! I was expecting just a minor improvement, so this is great.

santisoler (approved):

santisoler:
Thanks for applying those changes @mdtanker ! This is looking great.

Regarding the tests. Harmonica test functions live in `harmonica/tests`. Each test function should be named `test_...` so `pytest` can know that it should run it.

There are different types of tests that we can run. The most important ones in Harmonica are the ones that check if a computation is producing accurate results. These are critical for ensuring that our tools compute what they meant. In `harmonica/tests/test_prism.py` we have several test functions for the `prism_gravity` function. I would add a function in there that checks if the results when passing `progressbar=True` are the same as the ones obtained when passing `progressbar=False`.

The same type of function should be added to `harmonica/tests/prism_layer.py` but for the prism layer.

You'll see that the test functions usually have a bunch of decorators (those lines that start with `@`). They are intended to use some of the nice tools provided by `pytest` like parametrization (running the same test under different scenarios) or fixtures (create sample objects that test functions use as input, like sample computation points or a sample collection of prisms). You can learn more about them in `pytest` docs: https://docs.pytest.org/en/6.2.x/parametrize.html

Besides, you'll find that we often use two different decorators:
- `@pytest.mark.use_numba`
- `@run_only_with_numba`

The first decorator should be added to every test function that tests a code written in Numba, while the second one is intended to be used on those test functions that take too long to run if Numba JIT is disabled.
Don't worry too much about the second one for now, but on every test function that will be computing the gravity of prisms use the first one.

Besides those there are other type of tests we should run, which are more intended to guarantee that the user interface works as expected. For instance, we need to check that the `prism_gravity` function raises a `ImportError` if `numba_progress` is not installed. Check out how we do it for `pyvista`:
https://github.com/fatiando/harmonica/blob/6a30797fe179da63435be5204b7e384e38c7bde9/harmonica/tests/test_visualizations.py#L121-L128

When you run `make test` you'll get a detailed report of which tests passes, which fails and also which lines are being tested and which aren't. The percentage of code tested is called test coverage. We aim to hit 100% test coverage, although sometimes it isn't possible. But always remember that 100% test coverage doesn't mean good tests. So we prioritize test quality over test coverage.

Feel free to ask for help. We can also add code to this branch if you need, so don't feel obliged to write every possible test.

santisoler:
Yesterday I was running some benchmarks on this branch against the code we have in `main` to test if there was any performance hit on adding the progress bar.

Here I leave the script I used for benchmarking:

<details>

```python
import time
import numpy as np
import verde as vd
import harmonica as hm

region = (0, 100e3, -40e3, 40e3)
spacing = 1000

easting, northing = vd.grid_coordinates(region=region, spacing=spacing)

wavelength = 24 * spacing
surface = np.abs(np.sin(easting * 2 * np.pi / wavelength))

density = np.full_like(surface, 2700)

prisms = hm.prism_layer(
    coordinates=(easting, northing),
    surface=surface,
    reference=0,
    properties={"density": density},
)

region_pad = vd.pad_region(region, 10e3)
coordinates = vd.grid_coordinates(region_pad, spacing=spacing, extra_coords=1e3)

# Run forward modelling once so Numba can compile
print("\nRunning first time for compilation")
prisms.prism_layer.gravity(coordinates, field="g_z")

# Then benchmark it
print("\nStarting benchmark")

times = []
for i in range(5):
    print(f"Running iteration {i}")
    start = time.time()
    gravity = prisms.prism_layer.gravity(coordinates, field="g_z")
    end = time.time()
    times.append(end - start)

print(f"Results: {np.mean(times)} +/- {np.std(times)}")
```

</details>

I run it as it is with the code in `main` to have a reference, and with the code in this branch choosing `progressbar=True` and `progressbar=False`. The results I got were:

| Type of benchmark |  Time elapsed (s) |
|---|---|
| `main` branch | 8.3 ± 0.4 |
| `progressbar=True` | 8.2 ± 0.8 |
| `progressbar=False` | 8.4 ± 0.4 |

> These specific values depend on the hardware setting and your cpu load at the time of the benchmark

In conclusion, I don't see any performance hit, not even comparing this branch with `main`, and not even choosing `progressbar=True` or `progressbar=False`.

mdtanker:
@santisoler thanks for the help with the tests. I've added test which I believe accomplish what we want. `make test` results are below.

```
----------- coverage: platform linux, python 3.9.0-final-0 -----------
Name                                                                              Stmts   Miss Branch BrPart  Cover   Missing
-----------------------------------------------------------------------------------------------------------------------------
/Users/home/tankerma/harmonica/harmonica/_version.py                                  3      0      0      0   100%
/Users/home/tankerma/harmonica/harmonica/constants.py                                 2      0      0      0   100%
/Users/home/tankerma/harmonica/harmonica/datasets/sample_data.py                     52      0      0      0   100%
/Users/home/tankerma/harmonica/harmonica/equivalent_sources/cartesian.py             88      0     24      0   100%
/Users/home/tankerma/harmonica/harmonica/equivalent_sources/gradient_boosted.py      79      0     30      0   100%
/Users/home/tankerma/harmonica/harmonica/equivalent_sources/spherical.py             66      0     12      0   100%
/Users/home/tankerma/harmonica/harmonica/equivalent_sources/utils.py                 29      0     14      0   100%
/Users/home/tankerma/harmonica/harmonica/forward/_tesseroid_utils.py                162      0     66      0   100%
/Users/home/tankerma/harmonica/harmonica/forward/_tesseroid_variable_density.py      67      0     16      0   100%
/Users/home/tankerma/harmonica/harmonica/forward/point.py                           103      0     22      0   100%
/Users/home/tankerma/harmonica/harmonica/forward/prism.py                            98      2     50      0    99%   141-142
/Users/home/tankerma/harmonica/harmonica/forward/prism_layer.py                     108      0     26      0   100%
/Users/home/tankerma/harmonica/harmonica/forward/tesseroid.py                        68      0     26      0   100%
/Users/home/tankerma/harmonica/harmonica/forward/utils.py                            57      0      8      0   100%
/Users/home/tankerma/harmonica/harmonica/gravity_corrections.py                      11      0      0      0   100%
/Users/home/tankerma/harmonica/harmonica/io.py                                       70      0     42      0   100%
/Users/home/tankerma/harmonica/harmonica/isostasy.py                                 18      0      2      0   100%
/Users/home/tankerma/harmonica/harmonica/synthetic/surveys.py                        34      0      6      0   100%
/Users/home/tankerma/harmonica/harmonica/visualization/prism.py                      31      2     12      0    95%   14-15
-----------------------------------------------------------------------------------------------------------------------------
TOTAL                                                                              1146      4    356      0    99%

================================================================================== short test summary info ===================================================================================
FAILED ../harmonica/forward/tesseroid.py::harmonica.forward.tesseroid.tesseroid_gravity
================================================================== 1 failed, 281 passed, 121 skipped, 56 warnings in 26.22s ==================================================================
make: *** [Makefile:26: test_coverage] Error 1

```
It resulted in 1 failure, which was on harmonica.forward.tesseroid.tesseroid_gravity so I don't think that is related. The tests had 99% coverage for prism.py. It was missing lines 141-142, the except statement in the following:

```python
# Attempt to import numba_progress
try:
    from numba_progress import ProgressBar
except ImportError:
    ProgressBar = None
```

Here was my test for whether numba_progress was installed, not sure if I did this properly.

```python
@patch("numba_progress.ProgressBar", None)
def test_numba_progress_missing_error():
    """
    Check if error is raised when progresbar=True and numba_progress package 
    is not installed.
    """
    prisms = [
        [-100, 0, -100, 0, -10, 0],
        [0, 100, -100, 0, -10, 0],
        [-100, 0, 0, 100, -10, 0],
        [0, 100, 0, 100, -10, 0],
    ]
    densities = [2000, 3000, 4000, 5000]
    coordinates = [0, 0, 0]
    # Check if error is raised
    with pytest.raises(ImportError):
        prism_gravity(
            coordinates, prisms, densities, field="potential", progressbar=True
        )
```

harmonica/forward/prism.py
```diff
@@ -130,9 +135,29 @@ def prism_gravity(
                 + "mismatch the number of prisms ({})".format(prisms.shape[0])
             )
         _check_prisms(prisms)
+    # Attempt to import numba_progress
+    try:
+        from numba_progress import ProgressBar
+    except ImportError:
+        ProgressBar = None
```
santisoler, line 142: Might be better to move this lines outside the function right after we run all the imports in `prism.py`. This way we don't try to import `ProgressBar` on each function call and only do it once when importing harmonica.

harmonica/tests/test_prism_layer.py
```diff
@@ -422,3 +428,37 @@ def test_to_pyvista(dummy_layer, properties):
         assert pv_grid.array_names == ["density"]
         assert pv_grid.get_array("density").ndim == 1
         npt.assert_allclose(pv_grid.get_array("density"), layer.density.values.ravel())
+
+
+@pytest.mark.use_numba
+def test_progress_bar(dummy_layer):
```
santisoler, line 434: Same here, add a `@pytest.mark.skipif` decorator.

santisoler (changes requested):
Looking great @mdtanker!

The tesseroid test error is probably unrelated. If that's the case we will handle it on a separate PR, so don't worry.

It's ok not to achieve 100% of coverage in this case. The `try: import...` lines are hard to test: we would need to mimic a missing `numba_progress` at import time. It's a tricky problem and we don't get much out of it besides a not very meaningful 100%. So it's ok if we leave those lines untested. The test for the custom `ImportError` that `prism_gravity` raises is great though! I only have a small suggestion on it, check comment below.

As you can see now we are running CIs on this branch to check that everything passes. We would need to add `numba_progress` as a requirement for running the tests. Please edit the `env/requirements-test.txt` file and add it there. CIs are also complaining about code style, that one is easy: just run `make format` to autoformat the code to comply with PEP8.

Thanks again @mdtanker ! Let me know if you need any help!

harmonica/forward/prism.py
mdtanker: Opps I thought I already fixed that, thanks!

mdtanker: Actually when I move these lines outside the function to the top, the tests failed, and don't raise the `ImportError`. Any suggestions here?

mdtanker:
```
FAILED ../harmonica/forward/tesseroid.py::harmonica.forward.tesseroid.tesseroid_gravity
FAILED ../harmonica/tests/test_prism.py::test_numba_progress_missing_error - Failed: DID NOT RAISE <class 'ImportError'>
FAILED ../harmonica/tests/test_prism_layer.py::test_numba_progress_missing_error[numpy] - Failed: DID NOT RAISE <class 'ImportError'>
FAILED ../harmonica/tests/test_prism_layer.py::test_numba_progress_missing_error[xarray] - Failed: DID NOT RAISE <class 'ImportError'>
```

santisoler: Nice catch. The problem there is in the test function `@patch` decorator, not with the `prism_gravity` function.
What we want to mock is the `ProgressBar` variable that gets defined in `harmonica.forward.prism` and not the `ProgressBar` class of the `numba_progress` module.

So, in the test function replace the first argument of the `@patch` decorator with `"harmonica.forward.prism.ProgressBar"`. It should look like this:

```python
@patch("harmonica.forward.prism.ProgressBar", None)
def test_numba_progress_missing_error():
    """
    Check if error is raised when progresbar=True and numba_progress package
    is not installed.
    """
    prisms = [
        [-100, 0, -100, 0, -10, 0],
        [0, 100, -100, 0, -10, 0],
        [-100, 0, 0, 100, -10, 0],
        [0, 100, 0, 100, -10, 0],
    ]
    densities = [2000, 3000, 4000, 5000]
    coordinates = [0, 0, 0]
    # Check if error is raised
    with pytest.raises(ImportError):
        prism_gravity(
            coordinates, prisms, densities, field="potential", progressbar=True
        )
```

mdtanker: Great! Yeah that makes more sense. I'll update and commit now. Thanks

santisoler (approved):

santisoler:
@mdtanker This is looking great. I think this is ready to be merged. Is there something else you want to add? We can always change things later, no problem.

mdtanker:
> @mdtanker This is looking great. I think this is ready to be merged. Is there something else you want to add? We can always change things later, no problem.

Merge away! Thanks for all the help.

PR #321, opened 2022-06-06
Specify "spherical latitude" when describing coordinates of point masses

Esteban82:
Add "spherical" when describing the spherical latitude coordinate of point
masses in the user guide. This way we differentiate it from the "latitude"
geodetic coordinate.

I just add 'spherical' following the description of the coordinate system where 'latitude' is used for geodetic (NOT spherical) coordinates.

Is this correct? Now I have some doubts.


Esteban82:
> You're right that we should put "spherical" in there to distinguish it from the geographical latitude. I'm merging this.

Below, in the same file there was another mention of "latitude". I am not sure if it should be "spherical latitude".

santisoler:
> Below, in the same file there was another mention of "latitude". I am not sure if it should be "spherical latitude".

Nice catch. I'm changing it right away.

PR #324, opened 2022-06-13
Note that spherical and geodetic latitudes are equal in spherical ellipsoids

Esteban82:
Add sentence in "important".

"They would be the same if the reference ellipsoid were a sphere. "


santisoler:
Like it! Merging this! 🚀

PR #328, opened 2022-06-27
Simplify tests for upward derivative

santisoler:
**Description**

Ditch the precomputed sample grid file and add new fixtures for creating
a pair of sample sources, a sample grid and some gravity fields as
DataArrays. Remove `netcdf4` from testing dependencies.

**Relevant issues/PRs**

Related to #299


LL-Geo:
> @LL-Geo how does it looks now? Does it makes a good boiler plate for the test you're working on in #299 ?

Hey @santisoler , that looks very good for me! I'm thinking to add a little complexity to the test. It might be good to put a pair of sources with positive and negative density contrast and depth.

harmonica/tests/test_transformations.py
```diff
@@ -7,75 +7,135 @@
...
+
+@pytest.fixture(name="sample_sources")
+def fixture_sample_sources():
+    """
+    Define a pair of sample point sources used to build tests
+    """
+    points = [
+        [-50e3, 50e3],
+        [-50e3, 50e3],
+        [-20e3, -20e3],
```
LL-Geo, line 28: might be good to have one shallow source and one deep source?

harmonica/tests/test_transformations.py
```diff
@@ -7,75 +7,135 @@
...
+    ]
+
+    masses = [3.3e8, 2.9e8]
```
LL-Geo, line 31: A positive and negative mass might be good as well?

harmonica/tests/test_transformations.py
santisoler: Sure! Well thought!

harmonica/tests/test_transformations.py
santisoler: Agreed! You're right that the model should be a little bit more complex.

santisoler:
Thanks @LL-Geo for the review! Making the sample model more complex was a clever idea. I'm merging this!

PR #329, opened 2022-06-28
Replace sphinx-panels for sphinx-design

santisoler:
Refactor the home page of Harmonica docs to use sphinx-design grid and cards.
Add some icons to the cards and redesign the header to be actual text instead
of a png banner. Add cards for each geometric body in Forward Modelling index
page and reorganize its content.

**Relevant issues/PRs:**

Inspired by https://github.com/fatiando/boule/pull/112


leouieda:
Looks great, @santisoler! 


As a side note, the install seems to be failing because netcdf4 doesn't have compiled wheels on python 3.7. Is it necessary for the tests?

leouieda:
Sorry, clicked the close button by mistake.

santisoler:
> As a side note, the install seems to be failing because netcdf4 doesn't have compiled wheels on python 3.7. Is it necessary for the tests?

Good question and excellent timing. I added netcdf4 as requirement for testing because I created a sample 2d grid for the FFT tests. But with #328 we don't need it anymore.

PR #339, opened 2022-08-16
Add a function to ignore the tesseroid with zero density or volume

aguspesce:
When I am working on `tesseroid_layer` function, I had an error for tesseroid with zero volume. 
So I add a new checked in `tesseroid_gravity` that ignore the tesseroid with zero density or zero volume


aguspesce:
Hi @santisoler! 
I think that it is OK. Please, take a look at this!

PR #341, opened 2022-08-19
Replace PROJECT placeholder in changes.rst for "harmonica"

BenjMy:
Hey, Harmonica curious user here! 

I found out that some links in the [change.rst](https://github.com/fatiando/harmonica/blob/main/doc/changes.rst) return a 404 error. 

I manually replaced all the `<https://github.com/fatiando/PROJECT/pull/*>` by  `<https://github.com/fatiando/harmonica/pull/*>`. 
Is it the way to do or is there any automatic implementation behind it? anyway, I let you review the PR according to your standards ;)

Thanks for your awesome work!

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -14,42 +14,42 @@ Version 0.5.0
 
 Deprecations:
 
-- Add ``FutureWarning`` to ``isostasy_airy`` function warning of deprecation after next release (`#307 <https://github.com/fatiando/PROJECT/pull/307>`__)
-- Ditch soon-to-be deprecated args of equivalent sources grid method (`#311 <https://github.com/fatiando/PROJECT/pull/311>`__)
-- Remove deprecated ``point_mass_gravity`` function (`#310 <https://github.com/fatiando/PROJECT/pull/310>`__)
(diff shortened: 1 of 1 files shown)


leouieda (approved):
@BenjMy thanks for spotting this and submitting a patch! This must have slipped through the cracks in #335 

We have a checklist item with a regex that does this but we've been needing to script the changelog generation for a while now (based on the checklist in that issue). Some of it has to be manual, like the sorting and summarizing of changes, but most of that could be automated with a script.

leouieda:
@santisoler should we make a 0.5.1 release to publish this fix?

santisoler:
Nice catch @BenjMy! This bug is on me.

>We have a checklist item with a regex that does this but we've been needing to script the changelog generation for a while now (based on the checklist in that issue). Some of it has to be manual, like the sorting and summarizing of changes, but most of that could be automated with a script.

We should add another item to the release checklist that runs `sed -i 's/PROJECT/harmonica/'`, because this is not the first time that I forget to replace the `PROJECT` placeholder (the first time it happened, I think I catched it before merging).

> @santisoler should we make a 0.5.1 release to publish this fix?

Sure! We can add #342 to 0.5.1 as well.

I'm merging this. @BenjMy, feel free to add yourself to the `AUTHORS.md` if that's something you want.

PR #342, opened 2022-08-19
Fix equivalent sources figures in gallery examples

BenjMy:
Hi Fatiando team!

First I'm happy to see that the upward continuation is now live in v0.5.0. I was using it until now with the old Fatiando!

This PR is related to a documentation issue and specifically to the example in the [equivalent source page](https://www.fatiando.org/harmonica/latest/gallery/equivalent_sources/gradient_boosted.html#sphx-glr-gallery-equivalent-sources-gradient-boosted-py)

**Issue**: the plot of the upward continued data is empty.

**Small tweak**: add the `region = vd.get_region(coordinates)` line to the examples. This allows getting the right coordinates to grid.

Looking forward to using the upward continuation feature!


santisoler:
Thank you very much for solving this @BenjMy! As you might have seen the docs had a lot of new stuff in there and apparently we forgot to check the gallery. I'm merging this! 🚀 

Like I mentioned in #341, feel free to add yourself to the `AUTHORS.md` file.

PR #351, opened 2022-09-14
Fix typo

MGomezN:
**Relevant issues/PRs:**


leouieda:
Good catch! I just read that paragraph and didn't see it.

PR #356, opened 2022-10-26
Set lower bounds for supported dependency versions

leouieda:
Follow NEP29 and set lower bounds for the dependencies (all minor versions from the last 2 years). The only exception is xrft, which is new and likely requires v1.0 to work with Harmonica. Setup CI to run on lowest Python with lowest dependencies, highest Python with highest dependencies, and highest Python with optional dependencies. Use Dependente to extract the version information from setup.cfg instead of the custom export tool we had.

**Relevant issues/PRs:**


leouieda:
@santisoler could we make the Actions checks not required? In practice we do require them but it's always a pain when we update the actions themselves, even if just changing a title here or there.

santisoler (approved):

PR #359, opened 2022-11-02
Set v0.4.0 as the last with support for Python 3.6

santisoler:
In `doc/compatibility.rst` replace Harmonica v0.5.0 for v0.4.0 for the
latest version that supports Python 3.6. We dropped support for it on
v0.5.0.

See release notes for v0.5.0: https://github.com/fatiando/harmonica/releases/tag/v0.5.0

**Relevant issues/PRs:**

Related to #309 and #356


santisoler:
@leouieda, I just noticed that you put v0.5.0 in the version compatibility page. But I think it should be v0.4.0. Is it right?

leouieda:
Oops, that was my mistake. I think I saw that and thought "0.5 -> Python 3.6".

PR #362, opened 2022-11-24
Clean API making some submodules private

### LINKED ISSUE #314: Enforce non-public subpackages as private modules?
Right now Harmonica has a several subpackages (`forward`, `equivalent_sources`, `visualization`), but only some of them are actually intended to be public.
For instance, it's ok for users to do:

```python
from harmonica.visualizations import prism_to_pyvista
```

But not ok for them to do:
```python
from harmonica.equivalent_sources import EquivalentSources
```

That's why we import the `EquivalentSources` class in the `__init__.py` and encourage users to do:

```python
import harmonica as hm

hm.EquivalentSources(...)
```

Currently, there's nothing enforcing users to not do the latter, besides documentation. But if anyone uses autocompletion tools (like in JupyterLab), every public module is listed.

🚨 **Question:** *Should we enforce it by making private subpackages actually private?* 🚨

This is done by renaming directories and files and putting a heading underscore, e.g. `_equivalent_sources`.

---

I think we should make a decision around this before the first major release of Harmonica.

santisoler:
Make the `forward`, `equivalent_sources`, `gravity_corrections`, `isostasy` and `transformations` submodules as private ones. This will enforce the usage of public functions and classes allowing developers to refactor private ones without the risk of braking backward compatibility. It also cleans up the list of options available when using autocompletion in IDEs and JupyterLab, helping the discoverability of the functions and classes we have in Harmonica.

**Relevant issues/PRs:**
Fixes #314


santisoler:
The `io` module should also be private, but this is handled by #348. To avoid having to solve conflicts I'm leaving the `io` untouched here.

aguspesce (approved):

PR #363, opened 2022-11-24
Replace setup.py with PyPA "build"

santisoler:
Use the build package to generate source and wheel distributions.
This means we can remove the setup.py file entirely since it's not
needed. Update the GitHub Actions workflows to match.

**Relevant issues/PRs:**

Inspired in https://github.com/fatiando/verde/pull/371


leouieda (approved):
Looks great!

PR #366, opened 2022-11-25
Remove deprecated functions

### LINKED ISSUE #340: Deprecate EQLHarmonic and EQLHarmonicSpherical
In #255 we renamed the `EQLHarmonic` and `EQLHarmonicSpherical` classes and marked the old ones for future deprecation. We should have done so before v0.5, but it's no a big deal, we can do it now then. I will add this Issue to the `v0.6` milestone so we don't forget to solve it before the next release.

**To Do**

Remove the `EQLHarmonic` and `EQLHarmonicSpherical` classes code, remove any reference to them in the docs and any test functions that use them.

**Tip**

You can use `git grep "EQLHarmonic"` to easily check if the `EQLHarmonic` text is still present in any place within the repo.

aguspesce:
Remove  `EQLHarmonic` and `EQLHarmonicSpherical`.  

Fix #340

### DIFF
--- a/harmonica/__init__.py
+++ b/harmonica/__init__.py
@@ -10,7 +10,7 @@ from . import datasets, synthetic
 from ._version import __version__
 from .equivalent_sources.cartesian import EquivalentSources
 from .equivalent_sources.gradient_boosted import EquivalentSourcesGB
-from .equivalent_sources.spherical import EQLHarmonicSpherical, EquivalentSourcesSph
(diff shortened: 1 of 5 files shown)


santisoler:
Thanks @aguspesce! I'm merging this :rocket:

Codecov is complaining because now we have a lower test coverage percentage. This is due to the fact that we now have less code to test, but not because we added untested code.

PR #372, opened 2022-12-08
Fix typos in docstrings of `prism_layer.py`

mdtanker:
Found some typos in the docstrings.

**Relevant issues/PRs:**


mdtanker:
No worries! I also noticed the "suggest edits" button on the docs page doesn't seem to work anymore? So I just made this PR instead.

santisoler:
Thanks for noticing that. The reason why the button doesn't work is because it points to the generated .rst file for that particular page. But we build that file when we build the docs: the actual docstring is inside the `.py` file where the function/method/class is implemented. I don't know if Sphinx Book theme has a way to override the path to the file that should be edited.. @leouieda do you have any idea? (BTW, I noticed that Verde has the same issue).

PR #373, opened 2022-12-15
Add option to discard thin prisms of a layer when forward modelling

### LINKED ISSUE #371: Ignore prisms with user-defined volume/thickness threshold

As discussed in the meeting today ([notes](https://hackmd.io/@fatiando/development-calls-2022#2022-12-07)), it would be great to allow users to vary the threshold for prisms of low volume/thickness. Currently, as introduced in PR #334, prisms with no volume are ignored when `prism_gravity` is called to save some unnecessary computation. In my use case, there are many prisms that are thin, but not exactly equal to 0, and thus are contributing to the run time, but not affecting the results significantly. 

It would be helpful if users could change some threshold parameter to account for this. 

## Option 1: 
Currently, zero-volume prisms are defined by this [line](https://github.com/fatiando/harmonica/blob/9529b5ace1056128bce000d2f8b9e88dd2b9016c/harmonica/forward/prism.py#L242): 
```python
# Mark prisms with zero volume as null prisms
null_prisms = (west == east) | (south == north) | (bottom == top)
```
We could change this to something similar to: 
```python
null_prisms = (abs(west - east) <= ew_thres) | (abs(south - north) <= ns_thres) | (abs(bottom - top) <= thick_thres)
```
where `ew_thres`, `ns_thres`, and `thick_thres` all default to 0. This would allow users to vary if they wanted.

## Option 2: 
Suggested by @santisoler, we could also implement this within the `prism_layer.gravity()` function. I assume this could be implement the same as the Option 1 example above.

Yes I'm happy to attempt this!

mdtanker:
Add a new `thickness_threshold` parameter to the `prism_layer.gravity()` method
that allows the user to ignore prisms below a certain thickness. This is
implemented within a new `_discard_thin_prisms()`, function which is similar to
the `_discard_null_prisms()` function. Add tests that check whether the
calculated gravity is the same from manually removing prisms and removing
prisms based on a threshold, and whether providing no threshold still produces
the correct gravity.

I've added a parameter `thickness_threshold` to the `prism_layer.gravity()` function which allows the user to ignore prisms below a certain thickness. This is implemented within a new function `_discard_thin_prisms()`, which is similar to the function `_discard_null_prisms()` [here](https://github.com/fatiando/harmonica/blob/9529b5ace1056128bce000d2f8b9e88dd2b9016c/harmonica/forward/prism.py#L214). 

I've also added a test that checks:
*  whether the calculated gravity is the same from manually removing prisms and removing prisms based on a threshold        

*  whether providing no threshold still produces the correct gravity. 

This shouldn't affect performance if no threshold is supplied since the discard is skipped if `thickness_threshold is None`.

I'm not sure what the performance hit is when a threshold is supplied.

**Relevant issues/PRs:**

Fixes #371


harmonica/forward/prism_layer.py
```diff
@@ -334,6 +341,9 @@ def gravity(
             Name of the property layer (or ``data_var`` of the
             :class:`xarray.Dataset`) that will be used for the density of each
             prism in the layer. Default to ``"density"``
+        thickness_threshold : int or float
+            Minumum prism thickness needed to be to including in the forward
+            gravity calculation. Default to 0.
```
santisoler, line 346: Small edit on the text (and fix typo):

```suggestion
            Minimum prism thickness needed to be included in the forward
            gravity calculation. Prisms thinner that this threshold will be ignored.
            Default to 0.
```

harmonica/forward/prism_layer.py
```diff
@@ -469,3 +486,43 @@ def to_pyvista(self):
...
+        Array containing the boundaries of the prisms in the following order:
+        ``w``, ``e``, ``s``, ``n``, ``bottom``, ``top``.
+        The array must have the following shape: (``n_prisms``, 6), where
+        ``n_prisms`` is the total number of prisms.
+    density : 1d-array
+        Array containing the density of each prism in kg/m^3. Must have the
+        same size as the number of prisms.
+    thickness_threshold : int or float
+        Minumum prism thickness needed to be to including in the forward
+        gravity calculation. Default to 0.
```
santisoler, line 511:
```suggestion
        Minimum prism thickness needed to be included in the forward
        gravity calculation. Prisms thinner that this threshold will be ignored.
        Default to 0.
```

harmonica/forward/prism_layer.py
```diff
@@ -469,3 +486,43 @@ def to_pyvista(self):
...
+    -------
+    prisms : 2d-array
+        A copy of the ``prisms`` array that doesn't include the thin prisms.
+    density : 1d-array
+        A copy of the ``density`` array that doesn't include the density values
+        for thin prisms).
+    """
+    west, east, south, north, bottom, top = tuple(prisms[:, i] for i in range(6))
+    # Mark prisms with thickness <= threshold  as null prisms
+    thickness = abs(top - bottom)
```
santisoler, line 523: I think this private function should assume that `top >= bottom` so the `abs` shouldn't be needed. If for some reason a prism in the layer has `bottom < top` we have a bug and we need to catch it beforehand on a specific check.
```suggestion
    thickness = top - bottom
```

santisoler (changes requested):
Thanks @mdtanker for opening this PR! This is looking great!

I left a few little comments below, nothing too big. I think the only thing that we would like to improve before merging this are the tests. We are currently testing if the gravity fields of two layers (one that have discarded thin prisms and one that doesn't) generate the same gravity field. This test makes sense, but we are not actually testing if discarding prisms based on thickness actually works: assume that the function has a bug and it doesn't discard any prism, the gravity field would still be exactly the same and all tests will pass, but the bug will still be there.

Would you like to write an extra test that checks just the private `_discard_thin_prisms` function? Basically it should build a set of prisms with different thicknesses and a set of densities (would be better if they all have different densities), pass them to the `_discard_thin_prisms` function and check if the output doesn't contain the thin prisms and the density values are the correct ones: the ones that correspond with the thick prisms.

The advantage of writing private functions in a modular way is that it's way easier to test them independently of the higher level code that uses them. This is usually known as _unit testing_ where we test the smallest units of our codebase independently of the rest.

harmonica/forward/prism_layer.py
mdtanker: Ok this makes sense. I was thinking that sense users can provide a `reference` which is actually above `surface` I would take the absolute value, but I didn't realize that harmonica would have flipped them internally. I think I'll open a separate issue to discuss this flipping of `top` and `bottom` when `bottom` is above `top`.

mdtanker:
@santisoler thanks for reviewing so fast! Good catch with adding that additional test. I'm having trouble using the `_discard_thin_prisms()` function outside of `prism_layer.py` (i.e.  inside the `test_prism_layer.py`). I've tried `from .. import _discard_thin_prisms`, which seems to work for `prism_layer` but not this function.  

Also, I noticed most of the other functions in `prism_layer` are actually methods of the class `DatasetAccessorPrismLayer`. Would it make since to re-write this function as one of these methods?

santisoler:
Hi @mdtanker! 

> I'm having trouble using the _discard_thin_prisms() function outside of prism_layer.py (i.e. inside the test_prism_layer.py). I've tried from .. import _discard_thin_prisms, which seems to work for prism_layer but not this function.

When you are trying to import with `from .. import _discard_thin_prisms` you are asking to go up one directory in the package (two dots `..` mean "the parent directory" in Unix-like OSs and also in the import statements in Python) and import `_discard_thin_prisms` from there, but that function is not available at that level. Since our tree is:
```
├── harmonica
│   │ ...
│   ├── tests/
│   │   ├── test_prism_layer.py
│   │   ├── ...
│   ├── __init__.py
│   │ ...
```
when you run `from .. import ...` you are going to be able to import only the functions and classes that are listed in the `harmonica/__init__.py`. But since `_discard_thin_prisms` is a private function it won't be available from there.

The good thing is that you can always import private stuff, but you need to provide the whole route to them. In this case you can import the `_discard_thin_prisms` function with:
```python
from ..forward.prism_layer import _discard_thin_prisms
```

> Also, I noticed most of the other functions in prism_layer are actually methods of the class DatasetAccessorPrismLayer. Would it make since to re-write this function as one of these methods?

I'm glad you bring this question. While doing object-oriented designs is usually tempting to add a new method to the class, even if it's a private one. But this instinct can lead to bugs or even to repeat code that could be broadly used throughout the project in the future. Every time you need to decide if some function should be a method or a function ask yourself: is this function actually needing some attribute of the class? In this particular case the `_discard_thin_prisms` function only needs a set of prisms, their densities and the density threshold. It can be applied even to a general list of prisms: there's nothing that bounds this function to the prism layer class. So in my opinion this should be a private function for now.

By contrast, every other method of the prism layer is actually reading some attributes from `self` to return something new or modifying some subset of these attributes. That's why those are methods and not functions. But even those being methods is also an interface decision that could be discussed. It is better to run the forward of the prism layer with a method?
```python
result = layer.prism_layer.gravity(coordinates, field)
```
or with a dedicated function?
```python
result = prism_layer_gravity(coordinates, layer, field)
```
Probably not the discussion for this PR, but I just wanted to mention it since behind our designs we are always taking these types of decisions.

mdtanker:
@santisoler thanks for the help. I've added a separate test to check that `_discard_thin_prisms` works correctly. Let me know if there's anything else you want to be tested here! 

Thanks for the explanation on methods vs. functions! Very helpful to have that explained.

harmonica/forward/prism_layer.py
```diff
@@ -469,3 +487,44 @@ def to_pyvista(self):
...
+        ``w``, ``e``, ``s``, ``n``, ``bottom``, ``top``.
+        The array must have the following shape: (``n_prisms``, 6), where
+        ``n_prisms`` is the total number of prisms.
+    density : 1d-array
+        Array containing the density of each prism in kg/m^3. Must have the
+        same size as the number of prisms.
+    thickness_threshold : int or float
+        Minimum prism thickness needed to be included in the forward
+        gravity calculation. Prisms thinner that this threshold will be
+        ignored. Default to 0.
```
santisoler, line 513: Update this description as well:
```suggestion
    thickness_threshold : float
        Prisms thinner than this threshold will be discarded.
```

harmonica/forward/prism_layer.py
```diff
@@ -469,3 +487,44 @@ def to_pyvista(self):
...
+
+    Returns
+    -------
+    prisms : 2d-array
+        A copy of the ``prisms`` array that doesn't include the thin prisms.
+    density : 1d-array
+        A copy of the ``density`` array that doesn't include the density values
+        for thin prisms).
```
santisoler, line 521:
```suggestion
        for thin prisms.
```

harmonica/forward/prism_layer.py
```diff
@@ -469,3 +487,44 @@ def to_pyvista(self):
...
+    """
+    west, east, south, north, bottom, top = tuple(prisms[:, i] for i in range(6))
+    # Mark prisms with thickness <= threshold  as null prisms
+    thickness = top - bottom
+    null_prisms = thickness <= float(thickness_threshold)
```
santisoler, line 526: Considering the description of the `thickness_threshold` parameter, wouldn't be better to mark the discarded prisms that have `thickness < thickness_threshold`?
```suggestion
    # Mark prisms with thickness < threshold  as null prisms
    thickness = top - bottom
    null_prisms = thickness < thickness_threshold
```

Also, I suspect that the float casting is not needed.

santisoler (changes requested):
Thanks @mdtanker for adding that test! I left a few comments below, mainly about the docstrings. After those are fixed I think this should be ready to be merged.

mdtanker:
Thanks for those suggestions. I've added them, and a few other minor changes. Happy for you to merge whenever 👍

santisoler (approved):

santisoler:
Thanks @mdtanker! I think this is ready to be merged! 🚀

PR #376, opened 2023-01-31
Fix format indicator of README for PyPI

leouieda:
It was still set as rst so the PyPI page was wrongly formatted. Change to markdown to fix the issue.

**Relevant issues/PRs:**


santisoler:
Thanks @leouieda! Nice catch!

PR #377, opened 2023-02-10
Add examples for grid transformations

santisoler:
Add a User Guide page showcasing examples on how to apply grid transformations using the Lighting Creek Sill Complex magnetic data available in Ensaio. Also add gallery examples for the upward continuation, upward derivative and
reduction to the pole using the same dataset.

**Relevant issues/PRs:**

Implement examples for functions introduced in #299


RichardScottOZ:
so we need a vertical derivative example?

also, a standard way to show people how to handle an irregular area - e.g. a polygon with nodata outside it?

santisoler:
> so we need a vertical derivative example?

In this PR I added a gallery example for the upward derivative (vertical derivative) and also in the User Guide. They both are the same example of computing the upward derivative on the [Lightning Creek Sill Complex magnetic grid](https://www.fatiando.org/ensaio/latest/gallery/lightning-creek-magnetic.html#sphx-glr-gallery-lightning-creek-magnetic-py) we have in Ensaio.

> also, a standard way to show people how to handle an irregular area - e.g. a polygon with nodata outside it?

That's a nice idea! I think I will add it in this PR, but we can open an issue for it, for sure. I have a few magnetic grids available in [NRCAN](https://natural-resources.canada.ca/home) in mind for it. Do you have any prefer method to deal with such grids besides replacing the `nan`s with zeros and then masking them out?

RichardScottOZ:
Nothing really cleverer than that, no.

PR #378, opened 2023-02-12
Compute horizontal derivatives of grids using finite differences

santisoler:
Make functions `derivative_easting` and `derivative_northing` to compute the horizontal derivatives using central finite-differences through the `xarray.DataArray.differentiate` method by default. Add a `method` argument to let users choose if they want to use the FFT implementation or the finite differences one. This implementation mimics the one we had in the old `fatiando`.


santisoler:
@LL-Geo, I'm planning to merge this PR during this week. If you have the chance to take a look at it, that would be nice. Otherwise, don't worry, we can always change this code afterwards.

Since horizontal derivatives using finite differences in the grid are more accurate than their FFT counterpart, I think it might be better to offer the finite differences option as the default one. This was also the default behavior in the old [`fatiando` horizontal derivatives functions](https://legacy.fatiando.org/api/gravmag.transform.html#fatiando.gravmag.transform.derivx).

santisoler:
@LL-Geo, I'm merging this so I can start the release of Harmonica. Don't worry about the review. If you find something we should change, feel free to open an issue or a pr! Thanks again for all the work with the fft filters!

LL-Geo:
Just had a quick look! That looks great! 👍

PR #389, opened 2023-03-02
minor typo in eq sources parameter estimation docs

mdtanker:
Fixed a small typo I noticed in the docs. Just missed `v.0.6.0` :unamused:


LL-Geo:
Aha, Nice catch 😄

PR #391, opened 2023-03-03
Minor edits to grid transformations guides

RichardScottOZ:
Fix some words.

I think this might be the clearest highpass/lowpass example I have seen written anywhere!


santisoler:
Thanks @RichardScottOZ, nice **multiple** catches! 

> I think this might be the clearest highpass/lowpass example I have seen written anywhere!

🤩 

I'm merging this!

BTW thanks @Esteban82 for the review!

PR #394, opened 2023-03-14
Remove horizontal coordinates when ditching thin prisms

santisoler:
Removing the unused horizontal coordinates when discarding thin prisms of a prism layer. Since we only check the vertical boundaries, we can exclude the horizontal ones.


mdtanker (approved):
Yeah, this looks good!

santisoler:
Thanks @mdtanker!

PR #400, opened 2023-04-17
Forward models of prisms gravity fields with Choclo

santisoler:
Replace the prisms gravity kernels for Choclo functions. Extend the capabilities of the `prism_gravity` function allowing to compute all the gravitational acceleration and tensor components. Add tests that compare Harmonica results against dumb Choclo calls and a Laplacian equation test for the diagonal components of the tensor. Ditch symmetry tests that were already covered by Choclo. Improve docstrings. Raise warning if any computation point falls in a singular point for the tensor components. Add tests for these new checks and warnings.


harmonica/_forward/prism.py
```diff
@@ -95,6 +125,36 @@ def prism_gravity(
...
+    Non-diagonal tensor components aren't defined on edges normal to the
+    remaining direction of the tensor (e.g. `"g_en"` is not defined on edges
+    parallel to the upward direction).
+    The function returns :func:`numpy.nan` on every singular point.
+
+    The diagonal tensor components aren't defined on observation points that
+    belong to the faces normal to their direction (e.g. ``"g_zz"`` is not
+    define on horizontal faces): two different limits exist when approaching
+    from the inside and from the outside of the prism.
+    This functions returns the limit of these components while approaching from
```
LL-Geo, line 149:
```suggestion
    This function returns the limit of these components while approaching from
```

LL-Geo (commented):
That looks great @santisoler 👍

santisoler:
Thanks @LL-Geo! I just pushed a few fixes. We can ignore codecov failing: we have less code now, since was moved to Choclo, so codecov sees that we are running less coverage percentage, but we are actually testing everything we want to test. I'm merging this!

PR #401, opened 2023-04-18
Warn for prism gravity tensor on singular points

santisoler:
Check if the gravity tensor components of a prism are being computed at any singular point. Choclo will return a ``np.nan`` on those points, but it's worth warning the user about it before they get the results. Add tests that check if the warnings are being raised on those cases.

**Relevant issues/PRs:**

Merge after #400 is merged.

Await until https://github.com/fatiando/choclo/pull/30 is merged and released.


santisoler:
Closing this, since I merged this branch into #400

PR #369, opened 2022-11-26
Add magnetic field forward modelling of rectangular prisms

santisoler:
Add forward modelling functions that calculate the magnetic fields of rectangular prisms. One function efficiently computes the three components of the magnetic field, while the other one calculates just one component. The functions make use of Choclo's forward modelling functions, while providing a nice interface, sanity checks and parallelization through Numba. Add examples and tests.

**Relevant issues/PRs:**

Related to #406

PR #402, opened 2023-04-22
Convert magnetic vector to inclination and declination

aguspesce:
Create functions to convert the three components of magnetic vectors to the intensity, inclination a declination, and vice-versa. Add new functions to the API reference and add tests for them.


harmonica/_utils.py
```diff
@@ -0,0 +1,141 @@
...
+        Declination angle of the magnetic vector.
+        If ``degrees`` is True, then the angle is returned in degrees, else it's
+        returned in radians.
+
+    Examples
+    --------
+    >>> intensity, inc, dec = magnetic_vec_to_ang(1.5, 1.5, -2.12132)
+    >>> print(intensity, inc, dec)
+    2.999999757066657, 44.99999536031435, 45.00000000000001
+    """
```
santisoler, line 121: Since we are going to operate with arrays inside the whole function, we can convert the inputs as arrays from the start. Later on we would have to return floats if they were originally floats.
```suggestion
    """
    # Convert inputs into arrays
    magnetic_e = np.atleast_1d(magnetic_e)
    magnetic_n = np.atleast_1d(magnetic_n)
    magnetic_u = np.atleast_1d(magnetic_u)
```

harmonica/_utils.py
```diff
@@ -0,0 +1,141 @@
...
+    # Compute the intensity as a norm
+    vectors = np.vstack((magnetic_e, magnetic_n, magnetic_u)).T
+    intensity = np.linalg.norm(vectors, axis=1)
```
santisoler, line 124: I think it's nice to use the `norm` function, but in this case we are unnecessarily allocating a 2d array with all the values of the magnetic componets. I would just use the old:
```suggestion
    intensity = np.sqrt(magnetic_e**2 + magnetic_n**2 + magnetic_u**2)
```

harmonica/_utils.py
```diff
@@ -0,0 +1,141 @@
...
+    # Fill the masked values
+    inclination = inclination.filled(-np.sign(magnetic_u) * np.pi / 2)
+    declination = declination.filled(0)
+    # Convert to degree if needed
+    if degrees:
+        inclination = np.degrees(inclination)
+        declination = np.degrees(declination)
+    if len(intensity) == 1:
+        intensity = intensity[0]
+    return intensity, inclination[0], declination[0]
```
santisoler, line 141: We want to return the intensity and two angles as the same type: either all floats or all arrays, so I'm fixing this a little bit. BTW, if we are passing multiple magnetic field vectors, then we expect multiple inclination and declination angles, not only the first element.
```suggestion
    if len(intensity) == 1:
        intensity, = intensity
        inclination, = inclination
        declination, = declination
    return intensity, inclination, declination
```

harmonica/_utils.py
```diff
@@ -0,0 +1,141 @@
...
+    declination : float or array
+        Declination angle of the magnetic vector.
+        If ``degrees`` is True, then the angle is returned in degrees, else it's
+        returned in radians.
+
+    Examples
+    --------
+    >>> intensity, inc, dec = magnetic_vec_to_ang(1.5, 1.5, -2.12132)
+    >>> print(intensity, inc, dec)
+    2.999999757066657, 44.99999536031435, 45.00000000000001
```
santisoler, line 120: Might be better to round the values so doctest can pass regardless of floating point errors.
```suggestion
    >>> round(intensity)
    3
    >>> round(inclination)
    45
    >>> round(declination)
    45
```

santisoler: BTW, pytest has an option for handling doctest with floats, but it was struggling with these cases where we have numbers like `2.9999999999997`. Might be better to keep an eye on it: https://docs.pytest.org/en/7.1.x/how-to/doctest.html

harmonica/_utils.py
```diff
@@ -0,0 +1,141 @@
...
+    inc_rad = np.radians(inclination)
+    dec_rad = np.radians(declination)
+    # Calculate the 3 components
+    magnetic_e = intensity * np.cos(inc_rad) * np.sin(dec_rad)
+    magnetic_n = intensity * np.cos(inc_rad) * np.cos(dec_rad)
+    magnetic_u = -intensity * np.sin(inc_rad)
+    return magnetic_e, magnetic_n, magnetic_u
+
+
+def magnetic_vec_to_ang(magnetic_e, magnetic_n, magnetic_u, degrees=True):
```
santisoler, line 56: Could we increase the verbosity in these functions names? I agree that using full words make them too long, but something like this might be better:
```suggestion
def magnetic_vec_to_angles(magnetic_e, magnetic_n, magnetic_u, degrees=True):
```

santisoler (changes requested):
This is looking great @aguspesce! I made some suggestions for changes, let me know what do you think.

I think the tests might need some work. For example, we are not testing against `degrees=False` and the pytest fixtures could be a little bit more readable.
Besides, for some reason tests aren't catching the bug in the last line of `magnetic_vec_to_ang` (where we return only the first element of `inclination` and `declination`). This is a sign that tests should be improved.

LL-Geo (commented):
Looks awesome! Just some minor suggestions for the docstring. 
Also, should we make a note somewhere, that our coordinate is upward up, so we have a negative magnetic vector (magnetic_u) with downward magnetization? Sometimes the upward is down, which might be a little confusing for people, haha.

santisoler:
> Looks awesome! Just some minor suggestions for the docstring. Also, should we make a note somewhere, that our coordinate is upward up, so we have a negative magnetic vector (magnetic_u) with downward magnetization? Sometimes the upward is down, which might be a little confusing for people, haha.

Thanks for the review @LL-Geo! You're right that we should probably be more explicit about directions. For Harmonica I'm considering that both for the magnetic and magnetization vectors, the vertical component points **upwards**. That's why we are using `magnetic_u` and not `magnetic_z`.

> Sometimes the upward is down, which might be a little confusing for people, haha.

Sounds like an [Escher's painting](https://upload.wikimedia.org/wikipedia/en/a/a3/Escher%27s_Relativity.jpg) haha. Sometimes people have coordinate systems with vertical pointing down. We try to overcome that confusion by explicitly putting the word **upward**.

The tricky thing in these functions is that the inclination angle is defined as positive when the vector points downwards. So, the equations for transforming them are not exactly the ones in Blakely (1995): in there they use z->down.

harmonica/_utils.py
```diff
@@ -0,0 +1,141 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+#
+# This code is part of the Fatiando a Terra project (https://www.fatiando.org)
+#
+import numpy as np
+
+
+def magnetic_ang_to_vec(intensity, inclination, declination):
+    """
+    Convert intensity, inclination and declination angles of the magnetic field to a
```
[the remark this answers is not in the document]
santisoler: These functions could be use with either magnetization vectors or magnetic vectors. We could improve the docstrings to make it more general: `magnetic field or magnetization vector` could be better suited here.

LL-Geo:
> Sounds like an [Escher's painting](https://upload.wikimedia.org/wikipedia/en/a/a3/Escher%27s_Relativity.jpg) haha. Sometimes people have coordinate systems with vertical pointing down. We try to overcome that confusion by explicitly putting the word **upward**.
> 
> The tricky thing in these functions is that the inclination angle is defined as positive when the vector points downwards. So, the equations for transforming them are not exactly the ones in Blakely (1995): in there they use z->down.

I like that painting! hahaha. That's the world in my dream🤣

Yeh... 'Traditional' coordinates usually point vertically down. And always confuse people. One day I suddenly realised that the positive density generates negative gravity in SimPEG due to the definition of "positive up and down". hahaha...

santisoler:
> > Sounds like an [Escher's painting](https://upload.wikimedia.org/wikipedia/en/a/a3/Escher%27s_Relativity.jpg) haha. Sometimes people have coordinate systems with vertical pointing down. We try to overcome that confusion by explicitly putting the word **upward**.
> > The tricky thing in these functions is that the inclination angle is defined as positive when the vector points downwards. So, the equations for transforming them are not exactly the ones in Blakely (1995): in there they use z->down.
> 
> I like that painting! hahaha. That's the world in my dreamrofl
> 
> Yeh... 'Traditional' coordinates usually point vertically down. And always confuse people. One day I suddenly realised that the positive density generates negative gravity in SimPEG due to the definition of "positive up and down". hahaha...

Indeed! That's why we compute the `g_z` in Harmonica, but explicitly saying that it's the **downward** component, although our coordinate system has the vertical pointing upwards.

santisoler (approved):

santisoler:
@aguspesce I think this is ready to be merged after I took the liberty to make a few changes. Would you like to review it and share your thoughts?

Although it would be nice to have some example for these two functions, I think we can avoid including them here and just use these functions in other example that forward models some magnetic field. For example, the ones in #369 

Thanks a lot for the time you put into this!

aguspesce:
@santisoler and @LL-Geo thanks you for your collaboration on this and finished this code for me.

PR #408, opened 2023-05-11
Update gravity_disturbance_topofree.py

Esteban82:
At the beginnig of the scripts says: "The gravity disturbance is the observed gravity minus the normal gravity (boule.Ellipsoid.normal_gravity). The signal that is left is assumed to be due to the differences between the actual Earth and the reference ellipsoid." So, the value 0 has a special meaning. 

But, in the figure the hinge (white color) of the CPT is around 100. I think that it should be in 0.

To fix it, I propose to force the hinge at 0 (+h0).

### DIFF
--- a/examples/gravity_disturbance_topofree.py
+++ b/examples/gravity_disturbance_topofree.py
@@ -61,7 +61,7 @@ disturbance_topofree = disturbance - bouguer
 # Make a plot of data using PyGMT
 fig = pygmt.Figure()
 
-pygmt.grd2cpt(grid=disturbance_topofree, cmap="vik", continuous=True)
+pygmt.grd2cpt(grid=disturbance_topofree, cmap="vik+h0", continuous=True)
 
 title = "Topography-free (Bouguer) gravity disturbance of the Earth"
 


santisoler:
Nice catch @Esteban82! Now I'm thinking if we should use the `polar` cmap instead, as it's the one being used for the gravity disturbance example (https://www.fatiando.org/harmonica/latest/gallery/gravity_disturbance.html#sphx-glr-gallery-gravity-disturbance-py).

santisoler:
I just updated this branch after I merged #409 to `main`. Now tests should pass here as well.

Esteban82:
> we should use the `polar`

Probably yes for a paper. But for the gallery I think it is better to have figures with differents colors.

santisoler:
Thanks @Esteban82 for the contribution!

santisoler:
@Esteban82 I was checking the most recent build of the topo-free disturbance and noticed that the hinge is not set to zero: https://www.fatiando.org/harmonica/dev/gallery/gravity_disturbance_topofree.html#sphx-glr-gallery-gravity-disturbance-topofree-py

I was wondering if there was any change in pygmt that makes the hinge not be centered around zero even when using `cmap="vik+h0"`. Do you know what's going on?

Esteban82:
Hi Santi! Recently I had the same impression that something happened to the hinge. I will check it.

leouieda:
I've also been having this issue in my classes. I think it may be a GMT bug. When using `makcpt` the hinge works but not when letting GMT make the cpt automatically.

Esteban82:
I already reported the bug (https://github.com/GenericMappingTools/gmt/issues/8102).

Esteban82:
@santisoler the bug was solved in the master version.

santisoler:
Awesome response time! Thanks a lot @Esteban82!

I was also thinking that for PyGMT it would probably be nice to have some argument to set the hinge to zero, besides of adding `"+h0"` to the colormap. I think is hard to find out how to do it by reading the docs. But this should probably worth opening an issue on PyGMT.

Thanks again for taking care of this issue! Looking forward to see the next release of GMT and see the Harmonica docs updated.

PR #419, opened 2023-06-30
Rewrite check for point inside tesseroid with Numba

### LINKED ISSUE #417: Check for observartion point inside tesseroid eats too much memory
**Description of the problem:**

The tesseroid forward modelling function checks if no observation point falls inside any tesseroid through the  `_check_points_outside_tesseroids` function:
https://github.com/fatiando/harmonica/blob/b8f037e6d72b04cb0b4ff62bb12b63dd13e38808/harmonica/_forward/_tesseroid_utils.py#L405-L461

This functions makes use of Numpy arrays to run these checks, but this is super inefficient, since Numpy will try to allocate massive arrays of shapes (`n_coords`, `n_tesseroids`).

**Solution**
We should rewrite this function using Numba instead, making sure that we aren't allocating any additional array and the function just checks if observation points don't fall inside any tesseroids. We could also benefit from Numba parallelizations and run the first for loop in parallel (this should follow the user's choice set through the `parallel` argument in `tesseroid_gravity`).

As a quick design, we could do something like:
```python

@numba.jit(nopython=True, parallel=True)
def check_points_outside_tesseroids(coordinates, tesseroids):
    """..."""
    longitude, latitude, radius = coordinates[:]
    for i in numba.prange(longitude.size):
        for j in range(tesseroids.shape[0]):
            _check_point_inside_tesseroid(
                longitude[i],
                latitude[i],
                radius[i],
                tesseroids[j, 0],
                tesseroids[j, 1],
                tesseroids[j, 2],
                tesseroids[j, 3],
                tesseroids[j, 4],
                tesseroids[j, 5],
            )
            
@numba.jit(nopython=True)
def _check_point_inside_tesseroid(
    longitude, latitude, radius, west, east, south, north, bottom, top
):
    """Raise error if observation point falls inside tesseroid"""
    ...
    if ... :
        raise ValueError(...)
```

Check out https://github.com/fatiando/harmonica/blob/main/harmonica/_forward/prism.py for inspiration on how to code this.

**Temporary solution**

For the moment, the check can be bypassed by passing `disable_checks=True`, but it also disables other checks.

santisoler:
Rewrite the functions that checks if a computation point falls inside a tesseroid using Numba instad of Numpy. This greatly reduce the memory need of the Numpy functions, that required to build large matrices even for medium-size problems.

**Relevant issues/PRs:**

Fixes #417


harmonica/_forward/_tesseroid_utils.py
```diff
@@ -423,42 +423,95 @@ def _check_points_outside_tesseroids(coordinates, tesseroids):
...
+
+
+@jit(nopython=True)
+def _check_points_outside_tesseroids(coordinates, tesseroids):
     longitude, latitude, radius = coordinates[:]
-    west, east, south, north, bottom, top = tuple(tesseroids[:, i] for i in range(6))
+    conflicting = []
+    for i in range(longitude.size):
+        for j in range(tesseroids.shape[0]):
+            if _check_point_inside_tesseroid(
```
leouieda, line 455: @santisoler is it worth having a function here? The function itself is a single `if` and is called inside another `if`.

santisoler: Not at all. I coded this fairly quickly and I was prototyping a little bit with parallelization of the checks. I ultimately decided to keep them in serial. I've already merged the two functions into a single one.

santisoler:
@leouieda I think this is ready to be merge. If you would like to give it a second look, feel free to do so.

PR #438, opened 2023-09-22
Ditch setup.cfg and use only pyproject.toml

santisoler:
Remove `setup.cfg`. Move all project metadata and build instructions to `pyproject.toml`. Move `flake8` configuration to `.flake8`. Use `harmonica/_version.py` to store the generated version by `setuptools-scm`, and append the leading `v` to the version number directly in `harmonica/__init__.py`. Pin newer version of `setuptools` and `setuptools_scm`. Make use of `dependente==0.3.0` that adds support to packages that use `pyproject.toml` only.


leouieda (approved):
Looks great! Would be nice to get rid of `.flake8` at some point but 🤷🏾 

The reason we had a  `_version_generated` and a `_version` file was because of Pooch. When data were hosted here, we needed to import the `__version__` string and it caused a circular import if we fetched it from the base `__init__.py`. Now that we use Ensaio, we don't need it anymore.

PR #441, opened 2023-10-30
Add a finite-difference upward derivative

leouieda:
The `upward_derivative` function gains a `method` argument that defaults to `finite-diff`. The way this works is by upward and downward continuing the grid by half of the average grid spacing and then calculating a central difference approximation for the derivative. An advantage of this method is a lower sensitivity to noise than the FFT-based one. Even though the continuation is FFT-based, it's not very sensitive to high-frequency noise.

### DIFF
--- a/harmonica/_transformations.py
+++ b/harmonica/_transformations.py
@@ -7,6 +7,8 @@
 """
 Apply transformations to regular grids of potential fields
 """
+import numpy as np
+
 from .filters._filters import (
     derivative_easting_kernel,
     derivative_northing_kernel,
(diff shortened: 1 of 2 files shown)


leouieda:
Trying this out. We've had good experiences with it in [our magnetic microscopy paper](https://github.com/compgeolab/micromag-euler-dipole/blob/1ecf9d1aeff7dada8878957e82dc30224a2b79fe/code/micromag.py#L382). This was much better than the FFT-based one, particularly for Euler Deconvolution where the quality of the upward derivative is very important.

Implemented a test but it's still failing. Need to see if this is something I'm doing wrong in the code of if the results aren't as good as I thought.

santisoler:
@leouieda I think this PR just needs some little work to be ready, and I think it would be great to have it in the next release of Harmonica. Is that possible? What do you think?

leouieda:
@santisoler not sure I'll be able to finish this in time. I'll try but don't let this be a bottleneck for a release.

RichardScottOZ:
Much better in what sense for the layperson Leo?

PR #442, opened 2023-10-30
Use a text field for license in pyproject.toml

leouieda:
When using the "file" field type, PyPI putting the entire contents of the file into the license field and mislabeling it as MIT. Replace it with just the string for the license name and not the contents of the file.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,26 +1,22 @@
-[build-system]
-requires = ["setuptools>=61", "wheel", "setuptools_scm[toml]>=8.0.3"]
-build-backend = "setuptools.build_meta"
-
 [project]
 name = "harmonica"
 description = "Forward modeling, inversion, and processing gravity and magnetic data"
 dynamic = ["version"]
 authors = [
-  { name="The Harmonica Developers", email="fatiandoaterra@protonmail.com" },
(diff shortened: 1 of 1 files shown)


leouieda:
See https://test.pypi.org/project/harmonica/ for the incorrect output.

santisoler:
Nice catch @leouieda!

leouieda:
Ah no wait, there is also the classifier causing problems.

leouieda:
Can't fix it now, gotta run

leouieda:
Alright, done! Ready for a review. I made a few tweaks as well to put fields closer to other relevant fields.

pyproject.toml
```diff
@@ -1,26 +1,22 @@
...
 ]
 readme = "README.md"
-requires-python = ">=3.8"
-license = {file = "LICENSE.txt"}
+license = {text = "BSD-3-Clause"}
 keywords = ["geoscience", "geophysics"]
 classifiers = [
     "Programming Language :: Python :: 3",
-    "License :: OSI Approved :: MIT License",
+    "License :: OSI Approved :: BSD License",
```
santisoler, line 16: Ooops. Thanks for the fix!

santisoler:
I'm merging this. Thanks @leouieda!

PR #455, opened 2024-02-05
Use Dependabot to update GitHub Actions workflows

leouieda:
The Actions we use have pinned versions and it's almost impossible to keep up with new releases of them. Add configuration for Dependabot to track the Actions versions and send pull requests when new versions are available. This only works if version numbers are pinned instead of the commit hashes. I think the benefit of being updated out weight any risk associated with this.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -36,13 +36,6 @@ jobs:
       PYTHON: "3.11"
 
     steps:
-      # Cancel any previous run of the test job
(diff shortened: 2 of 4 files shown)


santisoler:
I noticed that dependabot replaces the hashes from commits with tags. Do we want to allow it to do so for actions that use our tokens?

If so, maybe we should update those comments saying we are actually pinning the commit hashes, right?

leouieda:
Hey Santi, I thought it would be fine. The only one we actually use is the Cancel Workflow one, which honestly, we can probably get rid of. The others are official PyPI actions, official Codecov actions, and official GitHub actions, so it shouldn't be a problem.

leouieda:
Now that I think about it, the only reason commits would be safer is if we audited the action to know that the particular commit doesn't do anything dangerous. We don't do this so I think the risk is pretty much the same.

leouieda:
Would you agree to remove the Cancel workflow actions?

santisoler:
I see. I think cancelling workflows is a nice feature, so we avoid using resources unnecessarily, but we can live with that.

I haven't look too much into it, but I think there's a way to cancel previous workflows through GitHub Actions API, specifically configuring the `concurrency`: https://www.meziantou.net/how-to-cancel-github-workflows-when-pushing-new-commits-on-a-branch.htm

I might need to research on that. But for now, feel free to remove it.

leouieda:
You can also do it manually. But yes, I think our development pace is slow enough that this isn't significant. I'll remove it here and elsewhere.

leouieda:
Done! Merging this in then.

PR #457, opened 2024-02-20
Fix broken ICGEM file loader test

leouieda:
The test was looking for a `UserWarning` with specific text that was changed in numpy. Fix it by making a much less specific regex and that will be compatible with both versions. Also fix a `FutureWarning` about accessing `xarray.Dataset` dimension sizes (should use `.sizes` instead of `.dims`).


santisoler:
Thanks @leouieda!

PR #468, opened 2024-03-05
Remove depth_type from EquivalentSources

### LINKED ISSUE #423: Remove the depth_type argument from equivalent sources

The `depth_type` argument makes explaining the `depth` argument challenging because it changes the meaning of `depth` entirely. It's also never something we recommend using. 

**Proposal:** Remove the `depth_type` argument and always use relative depths. This is almost always the best and other cases can be handled by passing custom `points`.

This breaks backwards compatibility but I think it's an improvement and we haven't reached 1.0 yet. It's also a rarely used argument so it won't impact many people. It's also never used in any of our docs. So I'm fine with breaking this without going through the hassle of emitting warnings first. **Any objections?**

Yes 💪🏾

santisoler:
Remove the `depth_type` argument from the `EquivalentSources` constructor. By default the point sources will be defined using a relative depth strategy. Removed the deprecated `relative_depth argument` from the constructor. Update the tests for the class and its methods.

Todo:

- [x] remove depth_type from gradient boosted equivalent sources and update their tests
- [x] remove depth_type from the docs (if any)
- [x] check if we need to update the spherical equivalent sources as well

**Relevant issues/PRs:**

Fixes #423


santisoler:
Co-authored by @indiauppal as well during a Fatiando Meeting in which we all worked together on this code.

PR #469, opened 2024-03-08
Use Burocrata to check and add license notices

leouieda:
Replaces the custom script in `tools`. Simplified the `Makefile` and GitHub Actions workflow to check style and format, instead of running individual tools separately. Ultimately, all of the formatting tools just need to be run if they fail, doesn't matter which.

**Relevant issues/PRs:**

Part of fatiando/community#140


santisoler (approved):
Looks awesome! Thanks @leouieda for these changes!

leouieda:
👍🏾

PR #474, opened 2024-03-12
Fix missing PyGMT images in user guide

### LINKED ISSUE #473: PyGMT figures are not being showed in the latest docs
**Page that contains the error or needs to be improved (paste link):** 

The PyGMT figures we generate while building the docs are not being correctly added to the `dev` versions of the Harmonica docs.

For example, compare the Gravity Disturbance page in:
- 0.6.0: https://www.fatiando.org/harmonica/v0.6.0/user_guide/gravity_disturbance.html
- dev: https://www.fatiando.org/harmonica/dev/user_guide/gravity_disturbance.html

I suspect the issue might come from a combination of PyGMT and jupyter-sphinx.

santisoler:
Add hidden cell in user guide examples configuring PyGMT to use `"notebook"` as display method. This way `jupyter-sphinx` is able to render the images in the generated HTML files. This issue should be fixed upstream in PyGMT.

**Relevant issues/PRs:**

Fixes #473


leouieda (changes requested):
Until there is a fix on the PyGMT side, I'd recommend adding this cell to the top of the files (after the imports):

```
.. jupyter-execute::
   :hide-code:

   # Needed so that displaying works on jupyter-sphinx and sphinx-gallery at
   # the same time. Using PYGMT_USE_EXTERNAL_DISPLAY="false" in the Makefile
   # for sphinx-gallery to work means that fig.show won't display anything here
   # either.
   pygmt.set_display(method="notebook")
```

This way the code works on notebooks and terminal (setting `method="notebook"` doesn't). And when there is a fix, this won't change anything.

santisoler:
I totally agree. That's a much better solution than this one. I'll revert the previous changes and then apply the one you suggested.

PR #472, opened 2024-03-12
Update versions of Sphinx and its plugins

santisoler:
Bump versions of Sphinx and its plugins in `environment.yml` and in `requirements-docs.txt`. Add CSS `dark-light` class to Harmonica logo in the `index.rst` to make its background transparent even in dark mode. Remove Harmonica logo in the left panel.

**Relevant issues/PRs:**

Similar changes to the ones applied in https://github.com/fatiando/choclo/pull/64

PR #476, opened 2024-03-12
Print xarray objects in user guide

santisoler:
Print `xarray.DataArray` and `xarray.Dataset` objects in the user guide pages instead of display them and letting Jupyter to render them as HTML objects. This is a patch to output cells not rendering properly on Sphinx dark mode.

**Relevant issues/PRs:**

This intends to patch the xarray issue mentioned in #472

### DIFF
--- a/doc/sg_execution_times.rst
+++ b/doc/sg_execution_times.rst
@@ -1,0 +1,85 @@
+
+:orphan:
+
+.. _sphx_glr_sg_execution_times:
+
+
+Computation times
+=================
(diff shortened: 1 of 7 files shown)


santisoler:
It's fine for me, @leouieda. I think we can go back to this PR before making the next Harmonica release and see if the problem was solved upstream. Otherwise we can apply this quick fix, release, and revert it in a future, when the displays look nice on dark theme. Does it make sense?

santisoler:
I'm closing this PR for now, but I'm not deleting the branch in case we need to get these changes back.

PR #477, opened 2024-03-13
Setup Trusted Publisher deployment to PyPI

leouieda:
Use this instead of the API tokens we used before. It's safer and easier to use because it doesn't required generating tokens and pasting them to GitHub.

**Relevant issues/PRs:**

Part of https://github.com/fatiando/community/issues/141


leouieda:
@santisoler you can delete the API tokens on PyPI and TestPyPI. I've already removed them from the repository.

santisoler:
Done!

PR #479, opened 2024-03-18
Fix inverted sign in upward derivative filter

### LINKED ISSUE #454: Inverted sign in upward derivative filter
**Description of the problem:**

The upward derivative returned has an inverted sign. It looks OK when inspecting plots because a lot of us are used to the sign convention of positive downwards. But since we use heights (and thus positive upward), the sign should be the opposite of what we expect. This is meant to improve consistency between small-scale and global data (which often has a "radial" component that is positive upward).

The fix should be relatively simple. Just add a minus sign to https://github.com/fatiando/harmonica/blob/main/harmonica/filters/_filters.py#L63

**Full code that generated the error**

```python
import harmonica as hm
import verde as vd
import numpy as np

model = [
    [35e3, 40e3, 50e3, 55e3, -6e3, -1e3],
]
magnetization = [
    hm.magnetic_angles_to_vec(20, -40, 15),
]
inc, dec = -40, 15
fe, fn, fu = hm.magnetic_angles_to_vec(1, inc, dec)
region = [0, 100e3, 0, 80e3]
coordinates = vd.grid_coordinates(region, spacing=500, extra_coords=1000)

be, bn, bu = hm.prism_magnetic(coordinates, model, np.transpose(magnetization), field="b")
anomalia = fe * be + fn * bn + fu * bu

grid = vd.make_xarray_grid(coordinates, anomalia, data_names="tfa", extra_coords_names="upward")
grid["d_up"] = hm.derivative_upward(grid.tfa)

fig, axes = plt.subplots(1, 2, layout="constrained", figsize=(10, 3))
grid.tfa.plot(ax=axes[0])
grid.d_up.plot(ax=axes[1])
```

**Full error message**

This produces the following result:

image

If we use positive upward, then the positive parts of the anomaly should be decreasing towards 0 (negative derivative) and the negative parts of the anomaly should be increasing towards 0 (positive derivative).

**System information**

* Operating system: Linux
* Python installation (Anaconda, system, ETS): Miniforge
* Version of Python: 3.11
* Version of this package: current main ([bbaf5a1](https://github.com/fatiando/harmonica/commit/bbaf5a191c7a3ccf9835a2d3b565be26a3500d26))
* If using conda, paste the output of `conda list` below: Not using conda.

<details>
<summary>Output of conda list</summary>
<pre>

PASTE OUTPUT OF CONDA LIST HERE

</pre>
</details>

YagoMCastro:
In order to apply the convention of positive upwards, this code:
1. Inverts signal of expected wavenumber at derivative_upward_kernel function;
2. Inverts signal of wavenumber at derivative_upward_kernel function;
3. Invert signal of upward derivative on test_transformations.py

**Relevant issues/PRs:**
Fixes #454


leouieda:
@YagoMCastro  thanks for the fix! Could you please add a description to the PR? Under "relevant issues" you should include `Fixes #454` so the issue will be linked here and closed automatically when this is merged.

harmonica/filters/_filters.py
```diff
@@ -61,7 +61,7 @@ def derivative_upward_kernel(fft_grid, order=1):
     k_northing = 2 * np.pi * freq_northing
     # Compute the filter for upward derivative in frequency domain
     da_filter = np.sqrt(k_easting**2 + k_northing**2) ** order
-    return da_filter
+    return -da_filter
```
leouieda, line 64: I'm not sure if the minus sign needs to be inside the `** order`. It would be good to test second derivatives as well to make sure.

harmonica/tests/test_transformations.py
```diff
@@ -258,7 +258,7 @@ def test_derivative_upward(sample_potential, sample_g_z):
         pad_width=pad_width,
     )
     # Calculate upward derivative and unpad it
-    derivative = derivative_upward(potential_padded)
+    derivative = -derivative_upward(potential_padded)
```
leouieda, line 261: This is the wrong place for the `-`. It should be further down in line 266. There we have:

```python
g_z = sample_g_z[trim:-trim, trim:-trim] * 1e-5  # convert to SI units
```

but it should be:

```python
g_up = -sample_g_z[trim:-trim, trim:-trim] * 1e-5  # convert to SI units
```

leouieda (changes requested):
Just need to fix the placement of the minus sign in the test. Could you also add a comment to the `test_derivative_upward_order2` function explaining that we don't need the minus sign there because the second derivative is positive both for z (downward) and upward derivatives?

leouieda (approved):
Great, well done!

PR #484, opened 2024-03-22
Extend support to Python 3.12

santisoler:
Bump minimum required Python version in CI. Bump Python version in `environment.yml`. Bump minimum version for some dependencies in `pyproject.toml`, requiring at minimum the ones that work on Python 3.12. Add classifier in `pyproject.toml` for Python 3.12.


santisoler:
Test fails because there's an issue with Verde and `pkg_resources` in Python 3.12. After that issue is fixed upstream in Verde, do we need to constrain Verde's version to that release?

santisoler:
Haha, apparently after closing https://github.com/fatiando/verde/pull/448, this PR was automatically closed. I'm reopening it since we do want to apply these changes despite the ones introduced in Verde.

pyproject.toml
```diff
@@ -35,7 +36,7 @@ dependencies = [
     "scikit-learn >= 0.24",
     "numba >= 0.52",
     "xarray >= 0.16",
-    "verde >= 1.7",
+    "verde >= 1.8.1",
```
leouieda, line 39: Are you sure this is required here? It's an issue with Verde only in Python 3.12 and Verde 1.7 should still work with previous Pythons.

santisoler: I'm not sure what to do here. If you're running Python 3.12 you **need** Verde >= v1.8.1.

leouieda: Yeah, you're right. I guess we should just bump the requirement here. It doesn't really have a huge impact and Verde 1.8 didn't break anything.

PR #487, opened 2024-04-02
Change default for window size in EquivalentSourcesGB

### LINKED ISSUE #425: Better default for window_size in EquivalentSourcesGB

The `window_size` in gradient-boosted equivalent sources currently defaults to 5 km. This would completely break for problems that have very large or very small areas. We used because we needed a default but this is not ideal.

A better default would be to estimate a square window where there will be about 5k data points on average. 5k data can fit on most computers RAM so it seems like a sensible default. Being conservative here means that we won't get memory errors from numpy in the majority of cases. In this case, the default would be `window_size=None` and in `.fit` we estimate a default value with:

```python
if self.window_size is None:
    area = (self.region_[1] - self.region_[0]) * (self.region_[3] - self.region_[2])
    ndata = data.size
    points_per_m2 = ndata / area
    window_area = 5e3 / points_per_m2
    self.window_size_ = np.sqrt(window_area)
else:
    self.window_size_ = self.window_size
```

And we use `self.window_size_` internally.

As with #424, I also think this is OK to break compatibility without going through the hassle of warning/deprecation. But will do it if others think it's needed.

Yes, but happy to let others do it since my time is limited.

indiauppal:
Change the default value for the `window_size` in the `EquivalentSourcesGB` constructor to `"default"`: window size will be estimated to locate approximately 5000 data points in each window. If `window_size="default"` and less than 5000 data points are used to fit the sources, a single window for all data and sources will be used. Check if the passed value of `window_size` is valid. Add a new `window_size_` attribute to the class that is set after fitting the sources. Update docstrings and extend tests.

**Relevant issues/PRs:**
Fixes #425

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -12,5 +12,4 @@ order by last name) and are considered "The Harmonica Developers":
 * [Santiago Soler](https://github.com/santisoler) - Department of Earth, Ocean and Atmospheric Sciences, University of British Columbia (ORCID: 0000-0001-9202-5317)
 * [Matthew Tankersley](https://github.com/mdtanker) - Antarctic Research Centre, Victoria University of Wellington, New Zealand (ORCID: 0000-0003-4266-8554)
 * [Leonardo Uieda](https://github.com/leouieda) - Universidade de São Paulo, Brazil (ORCID: 0000-0001-6123-9515)
-* [India Uppal](https://github.com/indiauppal) - The University of Liverpool, UK (ORCID:0000-0003-3531-2656)
 * [ziebam](https://github.com/ziebam) (not included in Zenodo)
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -12,4 +12,5 @@ order by last name) and are considered "The Harmonica Developers":
 * [Santiago Soler](https://github.com/santisoler) - Department of Earth, Ocean and Atmospheric Sciences, University of British Columbia (ORCID: 0000-0001-9202-5317)
 * [Matthew Tankersley](https://github.com/mdtanker) - Antarctic Research Centre, Victoria University of Wellington, New Zealand (ORCID: 0000-0003-4266-8554)
 * [Leonardo Uieda](https://github.com/leouieda) - Universidade de São Paulo, Brazil (ORCID: 0000-0001-6123-9515)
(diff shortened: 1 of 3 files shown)


indiauppal:
- [ ] Add a test to check window size default.

indiauppal:
- [ ] Raise an error for no. of points < 5e3
- [ ] Raise an error for no. of points = 5e3

santisoler:
This is starting to look great @indiauppal!

I'm leaving a few ideas after the meeting we had today:

1. We should probably extend the test function for less than 5000 points and test if the outputs match the ones for a single window. Something like:
    ```python
    data_windows, source_windows = eqs._create_windows(grid_coords)
    # The output lists should have a single element each (corresponding to the single window)
    assert len(data_windows) == 1
    assert len(source_windows) == 1
    # Check if all sources and data points are inside the window
    for coord in grid_coords:
        npt.assert_allclose(coord, coord[data_windows[0]])
    for coord in eqs.points_:
        npt.assert_allclose(coord, coord[source_windows[0]])
    ```
2. We could explore replacing those `np.arange`s with `slice`s. We only need those indices to slice the sources and coordinates arrays in the `self._gradient_boosting` method, so anything we could use to slice them that are more memory efficient should work. This is just a minor optimization, so it's not that critical for this PR.
3. Since now we have another attribute the `fit` method assigns (the `window_size_`), we should add it to the list of `Attributes` in the class docstring. Something like the think I wrote in #491 for the `depth_` argument:
    https://github.com/fatiando/harmonica/blob/55309c781e0059f78ced658c80d049ea222f31e4/harmonica/_equivalent_sources/cartesian.py#L142-L145
4. It would be nice to add a check for the `window_size` argument in the `__init__` method. Basically, we should raise an error if the passed value is a `str` and is not `"default"`. That's the only string value that should be valid. Check this out for inspiration: https://github.com/fatiando/harmonica/blob/55309c781e0059f78ced658c80d049ea222f31e4/harmonica/_equivalent_sources/cartesian.py#L176-L180

This is somewhat my personal wishlist for this PR, so feel free to assign me a few of these tasks if you want. As always, feel free to ask for help if you need it 🙂 

Looking forward to see this merged!

santisoler:
@indiauppal, I'm updating this branch after the fix I made for the failing Mac testing. Remember to run a `git pull` to sync your local repo with the latest change in this branch.

PR #488, opened 2024-04-08
Remove redundant urls in pyproject.toml

santisoler:
Reduce the number of urls in `pyproject.toml` by removing the ones that were redundant. Keep only links to `Documentation`, `Changelog`, `Bug Tracker`, and `Source Code`.

**Relevant issues/PRs:**

Related to https://github.com/fatiando/choclo/pull/77

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,13 +41,10 @@ dependencies = [
 ]
 
 [project.urls]
-"Homepage" = "https://www.fatiando.org/harmonica"
 "Documentation" = "https://www.fatiando.org/harmonica"
-"Repository" = "https://github.com/fatiando/harmonica.git"
 "Changelog" = "https://www.fatiando.org/harmonica/latest/changes.html"
 "Bug Tracker" = "https://github.com/fatiando/harmonica/issues"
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -41,13 +41,10 @@ dependencies = [
 ]
 
 [project.urls]
-"Homepage" = "https://www.fatiando.org/harmonica"
```
VascoSch92, line 44: This should be the link of the documentation

santisoler: Thanks for the comment @VascoSch92. In fatiando/choclo#77 we agreed that it's redundant to have Homepage and Documentation urls pointing to the same website. This is why I'm removing Homepage and leaving Documentation only.

I think it's nice to have Homepage and Documentation in case a project has its docs and website in different urls, like xarray (https://pypi.org/project/xarray). But for Harmonica it might be redundant since its website are its docs.

Do you agree?

VascoSch92: Ah ok ;)

Thanks for the clarification

leouieda:
Nice work, @santisoler!

PR #491, opened 2024-04-16
Change default value of depth in equivalent sources

### LINKED ISSUE #424: Better default for depth in equivalent sources

From chats with @indiauppal and @birocoles, I learned that Dampney (1969; the original equivalent source paper) advises using depths between 2.5 and 6 times the source spacing (assuming sources in a regular grid). For sources beneath data, like  we use here, a sensible default is 4.5 times the median distance between sources. Me and @indiauppal tested this default and it works well most of the time. Tweaking is always good but this would provide a more sensible default than the fixed value we currently use.

This would break backwards compatibility. I'm not too worried since our docs say that you should always set the `depth` by hand. So changing the default will likely make the interpolation better for the few people who were using the classes without following our advice. But I can do the warning dance if others think it's necessary (but it will take longer to implement).

Yes. But very happy to let someone else do it (😉 @indiauppal)

santisoler:
Change the default value of the `depth` argument in `EquivalentSources` to `"default"`. If this value is passed, then the depth of the sources is set as 4.5 times the mean distance between first neighbouring sources. If a numerical value is passed, this is the one that will be used. Introduce a new `depth_` attribute where the estimated/passed numeric value is stored. If `points` is passed, the `depth_` attribute is set to None. Add a test checking that the depth value is properly set. Merge the tests functions that checked the accuracy of equivalent sources using different dtypes.

**Relevant issues/PRs:**

Fixes #424


santisoler:
I should update some of the tutorials, so they make use of the default depth strategy. We can also include some information on why setting the depth to a numerical value and its consequences.

santisoler:
@indiauppal would you like to check this PR out?

harmonica/_equivalent_sources/cartesian.py
```diff
@@ -238,10 +263,14 @@ def _build_points(self, coordinates):
         """
         if self.block_size is not None:
             coordinates = self._block_average_coordinates(coordinates)
+        if self.depth == "default":
+            self.depth_ = 4.5 * np.median(vd.median_distance(coordinates, k_nearest=1))
```
santisoler, line 267: As discussed in #424, this should be the mean of the median distance
```suggestion
            self.depth_ = 4.5 * np.mean(vd.median_distance(coordinates, k_nearest=1))
```

harmonica/_equivalent_sources/cartesian.py
```diff
@@ -154,11 +168,16 @@ def __init__(
...
         points=None,
-        depth=500,
+        depth: float | str = "default",
         block_size=None,
         parallel=True,
         dtype="float64",
     ):
+        if isinstance(depth, str) and depth != "default":
+            raise ValueError(
+                f"Found invalid 'depth' value equal to '{depth}' ."
```
leouieda, line 178:
```suggestion
                f"Found invalid 'depth' value equal to '{depth}'. "
```

leouieda (approved):
Fixed a small typo. But otherwise, this is great!

harmonica/_equivalent_sources/cartesian.py
santisoler: Thanks!!

PR #502, opened 2024-05-08
Move codecov push to its own job in Actions workflow

santisoler:
Remove the push to codecov step from the `test` job into a new job that depends on the test job. Upload the coverage reports as artifacts after testing, and reuse the artifacts in the new job. Upload all coverage reports in a single push to Codecov to minimize the number of hits.

**Relevant issues/PRs:**

Related to fatiando/community#151

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -164,7 +164,7 @@ jobs:
 
   #############################################################################
   # Upload coverage report to codecov
-  codecov:
+  codecov-upload:
     runs-on: ubuntu-latest
     needs: test
     env:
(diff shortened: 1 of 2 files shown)


leouieda (approved):
Nice! For Harmonica this will work great but for other packages that have optional dependencies, the full coverage is only obtained when merging reports from multiple jobs. Then this won't work. Do you think it would be possible to upload artifacts for all of the jobs?

santisoler:
I think we can. We would just have to name each artifact differently and then download them all and pushed them all in a single job. Does it make sense?

leouieda:
👍🏾

santisoler:
Done! Now all runners upload the `coverage.xml` file as a separate artifact, and the `codecov-upload` job downloads them all and pushes them all together to Codecov in a single run (lowering the hits to their servers).

Let me know if you like it. I'm looking forward to try this out on a PR opened from a fork.

Issue #504, opened 2024-05-10
IGRF calculation


leouieda:
**Description of the desired feature:**

I'd like to implement calculation of the IGRF field. This would require the following components:

1. A way to calculate the associated Legendre functions $P_n^m$ without the Condon-Shortly phase and with Schmidt quasi-normalization. Equations for this can be found in https://www.gnu.org/software/gsl/tr/tr001.pdf
2. Reading the model coefficients from a file. 
3. Interpolating the model coefficients to the specified date. Not hard to do with Python's `datetime` module.

Since the model is only degree 13, we don't have to worry too much about the Legendre functions blowing up at larger orders. It would be nice to have the [Holmes and Featherstone (2002)](https://doi.org/0.1007/s00190-002-0216-2) method implemented later but it's not required for this job.

Design-wise, I have a few questions:

1. Should this be a function or a class? An `igrf` function would be very straight forward: calculate the 3-components at the given time and location. The location can be a float or array. But this won't produce an xarray grid. A class could have `IGRF.predict` and `IGRF.grid` like our equivalent-sources. But it won't have a `fit` since the model is already fit. In that case, `IGRF` could be an instance of a `SphericalHarmonics` class but I'm not sure I like that.
2. The calculations are done in geocentric spherical coordinates. But usually we'll have geodetic coordinates as inputs and the vector components output should be rotated to a local geodetic frame as well. At least, this is what web calculators and the NOAA and BGS programs calculate. Doing the conversion requires knowing the ellipsoid and so Boule comes in. So: should the function/class require an `bould.Ellipsoid` to be passed so it can do the conversions? Should it be optional (with `ellipsoid=None` assuming values in geocentric coordinates)? Or do we do everything in spherical and delegate the conversions entirely to Boule outside the IGRF function (this would require adding a vector rotation function to `boule.Ellipsoid`)?
3. Do we package the coefficient file or download it from NOAA? I'd be hesitant of NOAA changing their download location or simply changing the file when a new IGRF comes out. So maybe packaging it would be best.

I might break this down into multiple PRs adding different part as possibly private functions (Legendre and coefficient reading ones).

**Are you willing to help implement and maintain this feature?** 

Yes!

santisoler:
Excited to see this.

A few shallow comments after you ideas:

[...]

PR #505, opened 2024-05-10
Add associated Legendre function calculations

leouieda:
The numba-based functions calculate the associated Legendre functions (unnormalized, Schmidt normalized, and fully normalized) and their derivatives $\dfrac{\partial P_n^m}{\partial \theta}(\cos\theta)$. The derivatives functions can be used to calculate higher order derivatives as well. Values are tested against analytical solutions for the first 4 degrees as well as an identity for the Schmidt functions and the Legendre equation for the Schmidt and fully normalized functions. The values are returned as 2D numpy arrays. This is a bit of a waste of memory but it's much easier to use than storing these in sparse matrices or 1D arrays. But it should be acceptable since we're unlikely to ever go over ~2700 degree so the waste is in the order of a few Mb at worse.

**Relevant issues/PRs:**

Related to #504


leouieda:
@santisoler and @MarkWieczorek this is ready now. I added the Holmes and Featherstone scaling and tested until degree 2800. I was surprised that calculating the derivatives on rescaled functions works well. That's nice because it means we can keep them separate and use the derivative functions to calculate higher order derivatives (which would appear in gravity gradients).

harmonica/tests/test_legendre.py
```diff
@@ -0,0 +1,203 @@
...
+    p[3, 3] = 15 * np.sqrt(1 - x**2) ** 3
+    p[4, 0] = 1 / 8 * (35 * x**4 - 30 * x**2 + 3)
+    p[4, 1] = 5 / 2 * (7 * x**3 - 3 * x) * np.sqrt(1 - x**2)
+    p[4, 2] = 15 / 2 * (7 * x**2 - 1) * (1 - x**2)
+    p[4, 3] = 105 * x * np.sqrt(1 - x**2) ** 3
+    p[4, 4] = 105 * np.sqrt(1 - x**2) ** 4
+    return p
+
+
+def legendre_derivativeative_analytical(x):
```
santisoler, line 47: Fixing typo here
```suggestion
def legendre_derivative_analytical(x):
```

harmonica/tests/test_legendre.py
```diff
@@ -0,0 +1,203 @@
...
+        ),
+    ),
+    ids=["unnormalized", "schmidt", "full"],
+)
+def test_associated_legendre_function_derivative_analytical(func, norm, deriv):
+    "Check if the first few degrees match analytical expressions"
+    for angle in np.linspace(0, np.pi, 360):
+        x = np.cos(angle)
+        # Analytical expression
+        dp_analytical = legendre_derivativeative_analytical(x)
```
santisoler, line 146: And also here, where the function is being used.
```suggestion
        dp_analytical = legendre_derivative_analytical(x)
```

harmonica/tests/test_legendre.py
```diff
@@ -0,0 +1,203 @@
...
+    (
+        (associated_legendre, None),
+        (associated_legendre_schmidt, schmidt_normalization),
+        (associated_legendre_full, full_normalization),
+    ),
+    ids=["unnormalized", "schmidt", "full"],
+)
+def test_associated_legendre_function_analytical(func, norm):
+    "Check if the first few degrees match analytical expressions"
+    for angle in np.linspace(0, np.pi, 360):
```
santisoler, line 109: Just double checking that you actually meant to test with angles in $[0, \pi]$. I think it makes sense: if we extend it to $[0, 2\pi]$ we'll be duplicating the same tests, but just wanted to make sure of it.

harmonica/tests/test_legendre.py
```diff
@@ -0,0 +1,203 @@
...
+        35 * cos**4 - 15 * cos**2 + 15 * sin**2 - 105 * cos**2 * sin**2
+    ) / 2
+    dp[4, 2] = 105 * sin * cos**3 - 105 * cos * sin**3 - 15 * sin * cos
+    dp[4, 3] = 315 * cos**2 * sin**2 - 105 * sin**4
+    dp[4, 4] = 420 * sin**3 * cos
+    return dp
+
+
+def schmidt_normalization(p):
+    "Calculate the Schmidt normalization factor"
```
santisoler, line 77: Nit picking here
```suggestion
    "Multiply by the Schmidt normalization factor"
```

harmonica/tests/test_legendre.py
```diff
@@ -0,0 +1,203 @@
...
+    max_degree = p.shape[0] - 1
+    for n in range(max_degree + 1):
+        for m in range(n + 1):
+            if m > 0:
+                p[n, m] *= np.sqrt(2 * math.factorial(n - m) / math.factorial(n + m))
+
+
+def full_normalization(p):
+    "Calculate the full normalization factor"
```
santisoler, line 86: And here 😁 
```suggestion
    "Multiply by the full normalization factor"
```

harmonica/tests/test_legendre.py
```diff
@@ -0,0 +1,203 @@
...
+            for m in range(n + 1):
+                np.testing.assert_allclose(
+                    dp_analytical[n, m], dp[n, m], atol=1e-10, err_msg=f"n={n}, m={m}"
+                )
+
+
+def test_associated_legengre_function_schmidt_identity():
+    "Check Schmidt normalized functions against a known identity"
+    # Higher degrees than this yield bad results
+    max_degree = 2800
```
santisoler, line 165: Using the associated legendre functions with Schmidt normalization leads to inaccurate results for degrees larger than 2800? If that's so, I think we should document this in the function itself.

santisoler (changes requested):
Great work @leouieda! I think it will be great to have the spherical harmonics in Harmonica (redundancy required 😁), and this is step forward that.

I left some minor comments here. Let me know what do you think.

harmonica/tests/test_legendre.py
leouieda: Yep, that range is the valid range of colatitude, which is the angle used in spherical harmonics.

leouieda:
Thanks, @santisoler! What do you think?

harmonica/tests/test_legendre.py
santisoler: Great. Sorry I missed it.

PR #509, opened 2024-06-13
fix bug #439 with reduction_to_pole filter error

### LINKED ISSUE #439: Filter reduction_to_pole works only with "northing", "easting" coordinate names
My first issue on GitHub. Sorry if something is wrong:-)

I found that most filters are insensitive to xarray coordinates names; only the order is important.
But filter **reduction_to_pole** works only if coordinates have names "northing" and "easting". Otherwise error  "'freq_northing' is not a valid dimension or coordinate" appears. 

So maybe it could be useful to change row 475 in _filters.py 
from:
**da_filter.loc[dict(freq_northing=0, freq_easting=0)] = 0** 
to:
**da_filter.loc[{freq_northing.name:0, freq_easting.name:0}] = 0**

The current version generates dict with string keys: "freq_northing" and "freq_easting" while the suggested one generates keys according to the initial coordinate names.

zaarcvon:
**Relevant issues/PRs:**

Fixes #439 bug issue:  the reduction_to_pole filter produces an error when grid coordinate names are not "northing" and "easting". <br>
Now, it works regardless of grid coordinate names.

### DIFF
--- a/harmonica/filters/_filters.py
+++ b/harmonica/filters/_filters.py
@@ -472,7 +472,7 @@ def reduction_to_pole_kernel(
         magnetization_declination,
     )
     # Set 0 wavenumber to 0
-    da_filter.loc[dict(freq_northing=0, freq_easting=0)] = 0
+    da_filter.loc[{dims[0] : 0, dims[1] : 0}] = 0
     return da_filter
 
 


leouieda:
@zaarcvon thank you for the fix! Would you mind adding a quick test in that runs the reduction to the pole with a grid with dims that have another name? It could be a modification of this test https://github.com/fatiando/harmonica/blob/main/harmonica/tests/test_transformations.py#L615 that renames the dims of the test grid before doing RTP. Would be worth checking that the test fails with the code in `main` first.

zaarcvon:
Hi, @leouieda! I included a test function. I hope everything is done correctly. This is my first experience working with pytest

zaarcvon:
Hi, @leouieda . I see that some checks are failing. I understand one related to style and will fix it, but it's not clear what the code style/format check is really about.

santisoler:
Hi @zaarcvon. The two check that are failing are due to code style. The _format_ one is easy to solve, we just need to run black on the project. You can do it by running `make format` in your local repo (https://github.com/fatiando/community/blob/main/CONTRIBUTING.md#code-style). The _style_ one is telling us that flake8 is raising some warnings. You can check them out by clicking in [Details](https://github.com/fatiando/harmonica/actions/runs/9521345866/job/26286885817?pr=509), next to the failing check. In this particular case, flake8 is complaining about trailing whitespaces (whitespaces left after the last character of certain lines). When you run `make format`, black will take care of those as well.

santisoler:
Hi @zaarcvon! Today @YagoMCastro fixed the code style issues in your PR, so now we are ready to merge this. Thanks a lot for the contribution! 🚀

PR #510, opened 2024-06-14
Add `total_field_anomaly` function

### LINKED ISSUE #506: Function to compute the total field anomaly from magnetic field components

As discussed in today's meeting, it would be nice to include a function to compute the total field anomaly from the magnetic field components and with inclination and declination of the regional field. A function like such would be necessary to implement the ideas described in #426.

The function could be called `total_field_anomaly` and look something like:
```python
def total_field_anomaly(magnetic_field, inc, dec):
    b_e, b_n, b_u = magnetic_field
    ...
    tfa = ...
    return tfa
```

In order to get this done we should:
1. Implement the function in `harmonica/_utils.py`.
2. Make sure we are importing it in `harmonica/__init__.py` so it's accesible.
3. Add the new function to `doc/api/index.rst`.
4. Write tests for the function.

From what we talked in today's meeting, @indiauppal is willing to take the lead for this 🚀. I'll let her assign herself to this issue if she wants to.

indiauppal:
Add new function that computes the total field anomaly from the magnetic field components and the inclination and declination of the IGRF. Include the function in the API index and add tests for it.

**Relevant issues/PRs:**
Fixes #506


harmonica/_utils.py
```diff
@@ -148,3 +148,56 @@ def magnetic_vec_to_angles(magnetic_e, magnetic_n, magnetic_u, degrees=True):
...
+    where :math:`\mathbf{\hat{F}}` is the unit vector in the same direction
+    as the regional field.
+
+    Examples
+    --------
+    >>> tfa = total_field_anomaly([0, 0, -50e3], 90.0, 0.0)
+    >>> print(tfa)
+    50000.0
+    """
+    b_e, b_n, b_u = magnetic_field
```
VascoSch92, line 200: You could add a small check to control that the parameter `declination` is indeed in degree.

santisoler: Thanks for the suggestion @VascoSch92. It's hard to implement such checks. There's no way we can identify if the user passed an angle in radians or in degrees, besides making assumptions upon its value. I think documenting the argument properly, and keeping the consistency among the package (all inclination and declination angles, along with all longitude and latitude angles in Harmonica as expected in degrees) is enough to make users aware what is expected for these arguments.

VascoSch92: you are right :-)

PR #497, opened 2024-05-07
Drop support for Python 3.8

santisoler:
Bump minimal supported Python version to 3.8. Bump minimum versions of some dependencies (scipy, numba, pandas). Update the table of supported versions in the docs. Update test workflows to run 3.9 as the oldest version of Python.

PR #512, opened 2024-06-18
Drop macos x86 from Actions matrix

santisoler:
Remove the latest macos using x86 architecture from the GitHub Actions matrix we use for running tests. Since we dropped support for Python 3.8, we don't need to test Harmonica with oldest versions in macos runners with x86 cpus.

**Relevant issues/PRs:**

Related to #497


santisoler:
@leouieda, if we want to stop using x86 macos on CI, we should bump the minimum scikit-learn version, since PyPI doesn't provide binaries for sklearn 1.0.0 for arm64 macos. I think we can keep running test on macos x86 until we drop 3.9. What do you think?

PR #519, opened 2024-07-16
Add covariance, change fit data, more docs to Euler Deconvolution

leouieda:
The covariance matrix will be needed to filter solutions in the windowed approach. Change the fit input to receive all the data in a single tuple, like the coordinates. This is more compatible with the rest of Fatiando. Added the checks for fit input and standardizing array shapes so inputs can be n-dimensional. Copied over some of the docstrings from the legacy Fatiando implementation, which expands the theory and provides links to good papers to read.


Souza-junior:
@leouieda I'll take a closer look into it.

PR #521, opened 2024-07-23
Decorate tests for Legendre functions

santisoler:
Apply decorators to tests for the associated functions to compute the Legendre polynomials so we lower the time it takes to run the tests: mark the ones that test the Schmidt identity (running with high degrees) to require Numba JIT to be enabled, mark the other ones as `use_numba` so they are run with and without Numba JIT. Add low degree versions of the tests using the Schmidt identity to also check coverage with those.

### DIFF
--- a/harmonica/tests/test_legendre.py
+++ b/harmonica/tests/test_legendre.py
@@ -20,7 +20,6 @@ from .._spherical_harmonics.legendre import (
     associated_legendre_schmidt,
     associated_legendre_schmidt_derivative,
 )
-
 from .utils import run_only_with_numba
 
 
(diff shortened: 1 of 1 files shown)


santisoler:
I started this PR after @indiauppal and I noticed some of the tests taking too long. After merging #505, tests on CI went from 5mins to ~23mins. I'm aiming to reduce the run times back to around 5min.

santisoler:
I also noticed that the `harmonica/_spherical_harmonics/legendre.py` file is not being listed in the coverage report. I've checked locally and we are hitting 100% coverage on that file, but would be nice if we have it in the report too. I think as soon as we start using those functions in public modules, then they'll be included.

leouieda:
What? It wasn't taking that long in the PR was it?

santisoler:
It was, but we didn't notice:

- Actions in that PR: https://github.com/fatiando/harmonica/actions/runs/9469100858/job/26086983201
- Actions in main before that PR: https://github.com/fatiando/harmonica/actions/runs/9322173763/job/26093851286

leouieda:
Ah wow! I keep forgetting that Harmonica runs tests without numba compilation. That's something I really don't love but can't think of a way to get the full coverage without it.

leouieda (approved):
👍🏾 looks great!

PR #522, opened 2024-07-23
Solve seg fault while plotting with pyvista

santisoler:
Use qt version of `vtk` in `environment.yml`.


santisoler:
Docs are building properly now, so this PR is no longer needed. I'm closing it.

PR #537, opened 2024-10-29
Use the inc,dec to vector functions in RTP filter

leouieda:
The reduction-to-the-pole filter in the frequency domain did these conversions inside it before we had the dedicated functions. Replace that code with `magnetic_angles_to_vec` to simplify the code and remove the auxiliary function that was used. Revised the docstring and removed the conversion step since it's not done in this function anymore.


leouieda:
Was going through the code today for a class and noticed this. Always feels good to delete some code 🙂

Issue #538, opened 2024-10-30
GXF Reader Implementation


ThomasMGeo:
**Description of the desired feature:**

GXF is another Geosoft (now seequent) ASCII file format. It is found sometimes with USGS data releases like [here](https://pubs.usgs.gov/of/2000/ofr-00-0198/html/wyoming.htm). There is also this [gist](https://gist.github.com/jobar8/683483df605a906fb3da747b64627305) that I have found. 

**Are you willing to help implement and maintain this feature?** 
 Yes! I currently have functions that work for my needs, but I think this is complimentary to the grd file reader, and should be included in Harmonica. I am happy to do a PR, or help as needed. 
 
 One design thing I am concerned about is going from a GXF directly to xarray. I had to mess with the specific examples I had by hand to 'get it right' and wondering if that is best left to the user?

ThomasMGeo:
Here is my [gist](https://gist.github.com/ThomasMGeo/ddcc07b2e5f7b71a3d8cc698fba52534) with code.

ThomasMGeo:
@RichardScottOZ do you have a python GXF reader hiding somewhere?

leouieda:
@ThomasMGeo seems like a great thing to have in Harmonica! I don’t think I understand your reservation with going to xarray. From your gist, it seems like it would be easy to make the metadata CF-compliant and add it to a DataArray or Dataset. The main advantage is that everything else in Harmonica works with xarray and saving to another format like netCDF or Zarr is also trivial.

A PR would be very welcome! Even if just to get things started.

ThomasMGeo:
I will make it a data array in the PR, to just get it going. Thanks!

leouieda:
Would you be able to generate a small grid in that format so we can use for unit testing? It would have to be small so we can store it directly in the repository (like <1Mb).

RichardScottOZ:
> @RichardScottOZ do you have a python GXF reader hiding somewhere? 

No, sorry! Need a GDB reader though. Bound to have a few of the above in the archives I imagine.

PR #539, opened 2024-10-31
Create geosoft_gxf_io.py

ThomasMGeo:
GXF File Format IO is a common file type for grav/mag for USGS releases. It's an ASCII file type with variable header structures. 

On the dataset that I tested this on, it is upside down in the decoding to xarray. This is not impossible to fix, but I do like keeping the option for numpy and metadata output. 

I bet this could be optimized further, will try to find more datasets to test with (smaller the better!). Any comments/optimizations are welcome. 

**Relevant issues/PRs:**: This addresses #538 that has some more information about the format itself.

### DIFF
--- a/harmonica/_io/geosoft_gxf_io.py
+++ b/harmonica/_io/geosoft_gxf_io.py
@@ -1,0 +1,203 @@
+# Function to read in GXF files, provided by USGS Generally
+# No guarantee this will read in your specific GXF files
+
+# Written by Thomas Martin, Unidata
+# Inspired by this gist: https://gist.github.com/jobar8/683483df605a906fb3da747b64627305
+
+import numpy as np
+import xarray as xr
+
+from typing import Tuple, List, Dict, Union, Any
+
+def read_gxf_raw(infile: str) -> Tuple[List[str], Dict[str, str]]:
+    """
+    Read a GXF file and return raw data list and headers.
+    Following official GXF specifications from Geosoft.
+    
+    Parameters:
+    infile (str): Path to the GXF file
+    
+    Returns:
+    tuple: (data_list, headers)
+        - data_list: list of raw data strings
+        - headers: dictionary of GXF headers
+    """
+    # Read entire file
+    with open(infile) as f:
+        lines = [line.strip() for line in f.readlines()]
+    
+    # Create dictionary with headers and parameters
+    headers: Dict[str, str] = {}
+    data_list: List[str] = []
+    reading_data = False
+    
+    for i, line in enumerate(lines):
+        if not line:  # Skip empty lines
+            continue
+            
+        if reading_data:
+            data_list.append(line)
+        elif line.startswith('#'):
+            # Store the next non-empty line as the header value
+            key = line[1:]  # Remove the '#'
+            for next_line in lines[i+1:]:
+                if next_line and not next_line.startswith('#'):
+                    headers[key] = next_line
+                    break
+        
+        if line == '#GRID':
+            reading_data = True
+            
+    return data_list, headers
+
+def read_gxf(infile: str) -> Tuple[np.ndarray, Dict[str, Any]]:
+    """
+    Read a GXF file and return the grid data and metadata.
+    Following official GXF specifications from Geosoft.
+    
+    Parameters:
+    infile (str): Path to the GXF file
+    
+    Returns:
+    tuple: (grid_array, metadata)
+        - grid_array: numpy array containing the properly oriented grid
+        - metadata: dictionary containing all GXF parameters
+    """
+    # Get raw data and headers
+    data_list, headers = read_gxf_raw(infile)
+    
+    # Parse metadata with proper type conversion
+    metadata: Dict[str, Any] = {}
+    for key, value in headers.items():
+        try:
+            # Try converting to float first (handles scientific notation)
+            metadata[key] = float(value)
+            # If it's a whole number, convert to int
+            if metadata[key].is_integer():
+                metadata[key] = int(metadata[key])
+        except ValueError:
+            # If conversion fails, keep as string
+            metadata[key] = value.strip()
+    
+    # Convert data strings to numpy array
+    data_1d = np.array([])
+    for line in data_list:
+        # Handle both space and tab-delimited data
+        values = np.fromstring(line, sep=' ')
+        data_1d = np.concatenate((data_1d, values))
+    
+    # Get grid dimensions
+    nrows = int(headers['ROWS'])
+    ncols = int(headers['POINTS'])
+    
+    # Reshape to 2D array
+    grid_array = data_1d.reshape((nrows, ncols))
+    
+    # Handle dummy values
+    dummy_value = float(headers['DUMMY'])
+    grid_array[grid_array == dummy_value] = np.nan
+    
+    # Handle grid orientation based on SENSE parameter
+    sense = str(headers.get('SENSE', '1'))  # Default to '1' if not specified
+    if sense == '1':  # First point at bottom left of grid
+        grid_array = np.flipud(grid_array)
+    
+    # Add convenient grid parameters
+    metadata.update({
+        'nx': ncols,
+        'ny': nrows,
+        'x_inc': float(headers.get('PTSEPARATION', headers.get('XSEP', 1.0))),
+        'y_inc': float(headers.get('RWSEPARATION', headers.get('YSEP', 1.0))),
+        'x_min': float(headers.get('XORIGIN', 0.0)),
+        'y_min': float(headers.get('YORIGIN', 0.0))
+    })
+    
+    # Add projection information if available
+    if 'PRJTYPE' in headers:
+        metadata['projection'] = {
+            'type': headers['PRJTYPE'],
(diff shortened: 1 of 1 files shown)


ThomasMGeo:
I found a small file that works from here: https://pubs.usgs.gov/of/2000/ofr-00-0198/html/wyoming.htm

Specifically: mag5001_gxf.gz | 2000-04-24 16:50 | 57K 

Potential citation:

Kucks, R.P., and Hill, P.L., 2000, Wyoming aeromagnetic and gravity maps and data—A web site for distribution of data: U.S. Geological Survey Open-File Report 00-0198, https://pubs.usgs.gov/of/2000/ofr-00-0198/html/wyoming.htm

ThomasMGeo:
ToDo: Licensing of data file from USGS.

ThomasMGeo:
ToDo: Add Attribution to gist

ThomasMGeo:
For USGS, looks like we just need to add attribution as well: https://www.usgs.gov/faqs/are-usgs-reportspublications-copyrighted

ThomasMGeo:
@santisoler added some text about the USGS, the attribution to the gist was already there. Let me know if I need to do anything else! I probably won't edit this further until I have some more feedback.

ThomasMGeo:
@santisoler , is there anything else you want to see?

harmonica/_io/geosoft_gxf_io.py
```diff
@@ -0,0 +1,352 @@
...
+
+# Information about the GXF format found on a USGS Readme:
+
+"""
+1. Grid eXchange Format (*.gxf)
+
+GXF (Grid eXchange File) is a standard ASCII file format for
+exchanging gridded data among different software systems. 
+Software that supports the GXF standard will be able to import
+properly formatted GXF files and export grids in GXF format.
```
santisoler, line 15: I suspect that these paragraphs are coming from this PDF: https://pubs.usgs.gov/of/1999/of99-514/grids/gxf.pdf. Am I right?

Do we have permission to reproduce the same text here? Can we double check the license of those files? In doubt, I would remove these lines from this file, a reference should be enough. If we do have permission we can leave them, but I would make sure we are compliant with the license (reference the document, acknowledgement, include license text, etc?).

harmonica/_io/geosoft_gxf_io.py
```diff
@@ -0,0 +1,352 @@
+# Function to read in GXF files, provided by USGS Generally
+# No guarantee this will read in your specific GXF files
+
+# Written by Thomas Martin, Unidata
+# Inspired by this gist: https://gist.github.com/jobar8/683483df605a906fb3da747b64627305
```
santisoler, line 5: The gist is released under a BSD License, so we should acknowledge the author and specify the license.

```suggestion
# Inspired by this gist: https://gist.github.com/jobar8/683483df605a906fb3da747b64627305
# created by Joseph Barraud and made available under the BSD License.
```

harmonica/_io/geosoft_gxf_io.py
```diff
@@ -0,0 +1,352 @@
...
+    headers: Dict[str, str] = {}
+    data_list: List[str] = []
+    reading_data = False
+    
+    for i, line in enumerate(lines):
+        if not line:  # Skip empty lines
+            continue
+            
+        if reading_data:
+            data_list.append(line)
```
santisoler, line 188: From what I understood these lines are populating the `data_list` with the values of the grid. Since the number of lines for the grid are usually going to be way larger than the header lines, why not using `np.loadtxt` to read them? It performs much faster than appending lines to a list, it already returns an array in the right shape, and it also has some sanity checks (the number of elements per row are consistent) that we don't have to take care of.

I think we could start reading the header lines until we find the `#GRID` line, keep record of the number of rows for the header and stop the iteration. Then we can use `np.loadtxt` to read the rest of the file by passing the number of roads to `skiprows` (+/- 1, we need to figure out the correct indexing).

harmonica/_io/geosoft_gxf_io.py
```diff
@@ -0,0 +1,352 @@
...
+    # Convert data strings to numpy array
+    data_1d = np.array([])
+    for line in data_list:
+        # Handle both space and tab-delimited data
+        values = np.fromstring(line, sep=' ')
+        data_1d = np.concatenate((data_1d, values))
+    
+    # Get grid dimensions
+    nrows = int(headers['ROWS'])
+    ncols = int(headers['POINTS'])
```
santisoler, line 240: In lines https://github.com/fatiando/harmonica/pull/539/files#diff-5ac31ec9d8d11b47ff39f59bf25e1cad43484857580dd7b6c65cddd8e0f504cfR219-R229 we populate the `metadata` dictionary with parsed values (float, int, stripped str). Isn't it better to use `metadata` to get the number of rows and columns, instead of having to parse them again as ints?

That being said, maybe we want the keys in `metadata` to be lowercase (as you do in following lines with `nx`, `ny`, `x_inc`, etc.).

harmonica/_io/geosoft_gxf_io.py
```diff
@@ -0,0 +1,352 @@
...
+        print("\n=== Projection Information ===")
+        print(f"Type: {metadata['projection']['type']}")
+        print(f"Units: {metadata['projection']['units']}")
+        
+        params = metadata['projection']['parameters']
+        if any(params.values()):
+            print("\nProjection Parameters:")
+            for key, value in params.items():
+                if value is not None:
+                    print(f"{key}: {value}")
```
santisoler, line 308: I think this function is nice for debugging the reader. But I hardly see users using it (we would need to expose it as well). Instead, if `read_gxf` returns a DataArray, then we don't need this function: `xarray` can nicely display all the data.

What do you think? Is there any special reason to include this function?

santisoler (commented):
Thanks @ThomasMGeo for opening this PR!

I took a quick look at it and I have the following comments.

I'm not sure what's the benefit by not making the `read_gxf` to always return an `xarray.DataArray`. Once we have the DataArray, it's trivial to get the underlying Numpy array and the metadata as a dict. And by doing so, we only need to add one public function to Harmonica (`read_gxf`), instead of `read_gxf` and `gxf_to_xarray` (which has only a single use). In my opinion we should make `read_gxf` to behave similarly to the other readers in Harmonica and return a `DataArray`.

BTW, we need to remember to add any new public function to `harmonica/__init__.py` and to the API reference in `doc/api/index.rst`.

One little thing to note is that the docstrings should follow [numpydoc's style](https://numpydoc.readthedocs.io/en/latest/format.html), so they can be nicely rendered in Sphinx.

I left some more comments below.

Let me know what do you think. We can iterate over it later on! And feel free to ask for help if you need so.

harmonica/_io/geosoft_gxf_io.py
ThomasMGeo: I am happy to remove it.

ThomasMGeo:
Hi @santisoler , thanks for the review, I just want to touch on this few comments before I dig into the rest of the issues.

`I'm not sure what's the benefit by not making the read_gxf to always return an xarray.DataArray`

I am happy to have it be a very plain DataArray without the CRS (usually the CRS is not given as a code in the files that I have seen). For some of my work, I have had to do some weird warping to get them to match, but this might be rare and isolated. 

The get_grid_info was just used to figure out the projections (again, these are not standard CRS's with modern EPSG codes from what I have seen). So my general workflow would be to load it as a numpy array, look at that output, and then load it into xarray with all of the EPSG codes attached. 

Let me know what you think is best, and I can tackle it!

santisoler:
Hi @ThomasMGeo. I see your point. Since the gxf files don't have a strict specification for setting the CRS, I think we should just add to the `attrs` in the `xarray.DataArray` whatever information we can find in the header of the file. If that information is enough to determine the CRS, that's cool. But if it's not, too bad. I don't think there's nothing the reader function can do to solve that, am I right?

I do like having the information about the projection in the metadata. A nested dict (like the one you already implemented) should be ok (we might want to make sure that that structure follows the CF metadata conventions).

Another thing we were discussing today during the meeting is to use `easting` and `northing` for the x and y directions. But use `x` and `y` in case the grid is rotated (very much like the `read_oasis_grd` function does).

ThomasMGeo:
Awesome, that all makes sense. Let me work on it for a bit and get back to you. Thanks for giving it a look and for all the feedback! :)

harmonica/_io/geosoft_gxf_io.py
ThomasMGeo: We can drop this!

harmonica/_io/geosoft_gxf_io.py
```diff
@@ -0,0 +1,395 @@
...
+    tuple: (grid_array, metadata)
+        - grid_array: numpy array containing the properly oriented grid
+        - metadata: dictionary containing all GXF parameters
+    """
+    # First, read header efficiently until #GRID
+    headers: Dict[str, str] = {}
+    grid_start_line = 0
+    
+    with open(infile, 'r') as f:
+        lines = f.readlines()
```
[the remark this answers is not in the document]
ThomasMGeo: This should be fixed!

PR #552, opened 2025-02-19
Fetch Ensaio data from GitHub in CI

leouieda:
Set the `ENSAIO_DATA_FROM_GITHUB=true` environment variable in the docs build so data are fetched from GitHub instead of Zenodo. Helps reduce the amount of times we hit Zenodo for data downloads.


santisoler:
Thanks @leouieda!

PR #556, opened 2025-04-16
Add IGRF calculation

leouieda:
Add a class to calculate the IGRF14 field. Uses Pooch to download IGRF14 from Zenodo. The class takes a date to instantiate and then calculates the field for that data through `predict` and `grid` methods. Inputs and outputs are in geocentric spherical coordinates by default but can be changed to geodetic if a Boule ellipsoid is given to the constructor.

**Relevant issues/PRs:** #504

### DIFF
--- a/harmonica/_spherical_harmonics/igrf.py
+++ b/harmonica/_spherical_harmonics/igrf.py
@@ -1,0 +1,77 @@
+"""
+Calculation of the IGRF magnetic field.
+"""
+import pathlib
+import numpy as np
+import numba
+import xarray as xr
+from . import legendre
+from .._version import __version__
+
+import pooch
+import boule
+
+
+def fetch_igrf13():
+    """
+    """
+    path = pooch.retrieve(
+        url="doi:10.5281/zenodo.11269410/igrf13coeffs.txt",
+        path=pooch.os_cache("harmonica"),
+        known_hash="md5:e2e6e323086bde2dd910bb87d2db8532",
+    )
+    return path
+
+
+def load_igrf(path):
+    """
+    Load the IGRF Gauss coefficients from the given file
+    """
+    g, h, years = None, None, None
+    return g, h, years
+
+
+def interpolate_coefficients(date, g, h, years):
+    """
+    Interpolate the coefficients to the given date.
+    """
+    g_date, h_date = None, None
+    return g_date, h_date
+
+
+class IGRF13():
+    """
+    13th generation of the International Geomagnetic Reference Field
+    """
+
+    def __init__(self, date, ellipsoid=boule.WGS84):
+        self.date = date
+        path = fetch_igrf13()
+        g, h, years = load_igrf(path)
+        self._g, self._h = interpolate_coefficients(date, g, h, years)
+        self.reference_radius = 6371.2e3  # meters
+
+    @property
+    def dipole_moment(self):
+        """
+        Dipole moment of the Earth on a geocentric Cartesian system
+        """
+        return mx, my, mz
+
+    def predict(self, coordinates, field="b"):
+        """
+        Calculate the IGRF magnetic field at the given coordinates
+        """
+        latitude = coordinates[1]
+        longitude, latitude_sph, radius = self.ellipsoid.geodetic_to_spherical(
+            *coordinates
+        )
+        longitude = np.radians(longitude)
+        colatitude = np.radians(90 - latitude_sph)
+
+
+        return magnetic_field
+
+
+@numba.jit(nopython=True)
+def _cal
--- a/harmonica/_spherical_harmonics/igrf.py
+++ b/harmonica/_spherical_harmonics/igrf.py
@@ -2,14 +2,15 @@
 Calculation of the IGRF magnetic field.
 """
 import pathlib
-import numpy as np
-import numba
-import xarray as xr
-from . import legendre
-from .._version import __version__
 
-import pooch
 import boule
+import numba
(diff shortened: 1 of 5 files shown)


leouieda:
Running a benchmark against ppigrf:

```
Grid size: (594, 1201) = 713394
Harmonica: 0.156 +- 0.447 s 
ppigrf: 9.998 +- 0.163 s 
Harmonica is 63.89 times faster.
```

The Harmonica results include the numba compilation, hence the large std. Results are median times. This is the code for the benchmark:

```python
import numpy as np
import time
import datetime
import ppigrf
import harmonica as hm
import verde as vd


date = datetime.datetime.fromisoformat("2024-01-20")
coords = vd.grid_coordinates((0, 360, -89, 89), spacing=0.3, extra_coords=0)
print(f"Grid size: {coords[0].shape} = {coords[0].size}")

times_harmonica = []
for i in range(10):
    start = time.time()
    igrf = hm.IGRF14(date)
    b_harmonica = igrf.predict(coords)
    times_harmonica.append(time.time() - start)
median_harmonica = np.median(times_harmonica)
print(f"Harmonica: {median_harmonica:.3f} +- {np.std(times_harmonica):.3f} s ")

times_ppigrf = []
for i in range(5):
    start = time.time()
    b_ppigrf = ppigrf.igrf(*coords, date)
    times_ppigrf.append(time.time() - start)
median_ppigrf = np.median(times_ppigrf)
print(f"ppigrf: {median_ppigrf:.3f} +- {np.std(times_ppigrf):.3f} s ")

print(f"Harmonica is {median_ppigrf / median_harmonica:.2f} times faster.")

# Check that all components are close for both software
np.testing.assert_allclose(b_harmonica, [c[0] for c in b_ppigrf], rtol=0.001)
```

leouieda:
Here's an updated benchmark against ppigrf and SHTools:

```
Grid size: (601, 1201) = 721801
Harmonica       : 0.158 +- 0.451 s 
Harmonica (grid): 0.075 +- 0.315 s 
SHTools         : 0.015 +- 0.125 s 
ppigrf          : 10.253 +- 0.130 s 
```

Again, what's reported is the median time. So we're way faster than ppigrf but SHTools is still king (as expected). I'm actually surprised that we're not that much slower than SHTools but I think that'll change drastically if we do higher degree models. Still, the grid implementation is able to but down the time in half by avoid repeat computations. 


This is the benchmark script:

```python
import numpy as np
import warnings
import time
import datetime
import ppigrf
import harmonica as hm
import verde as vd
import pyshtools as psh
import boule as bl


# ppigrf doesn't handle the poles very well
warnings.filterwarnings("ignore")

date = datetime.datetime.fromisoformat("1900-01-01")
n = 600
region = (0, 360, -90, 90)
shape = (n + 1, 2 * n + 1)
height = 0
coords = vd.grid_coordinates(region, shape=shape, extra_coords=height)
print(f"Grid size: {coords[0].shape} = {coords[0].size}")

times_harmonica = []
for _ in range(10):
    start = time.perf_counter()
    igrf = hm.IGRF14(date)
    b_harmonica = igrf.predict(coords)
    times_harmonica.append(time.perf_counter() - start)
median_harmonica = np.median(times_harmonica)
print(f"Harmonica       : {median_harmonica:.3f} +- {np.std(times_harmonica):.3f} s ")

times_harmonica_g = []
for _ in range(10):
    start = time.perf_counter()
    igrf = hm.IGRF14(date)
    b_harmonica_g = igrf.grid(height, region, shape=shape)
    times_harmonica_g.append(time.perf_counter() - start)
median_harmonica_g = np.median(times_harmonica_g)
print(
    f"Harmonica (grid): {median_harmonica_g:.3f} +- {np.std(times_harmonica_g):.3f} s "
)

times_shtools = []
for _ in range(10):
    start = time.perf_counter()
    igrf = psh.datasets.Earth.IGRF_13(year=date.year)
    b_shtools = igrf.expand(
        a=bl.WGS84.semimajor_axis,
        f=bl.WGS84.flattening,
        lmax=(n - 2) / 2,
        lmax_calc=13,
    )
    times_shtools.append(time.perf_counter() - start)
median_shtools = np.median(times_shtools)
print(f"SHTools         : {median_shtools:.3f} +- {np.std(times_shtools):.3f} s ")

times_ppigrf = []
for _ in range(5):
    start = time.perf_counter()
    b_ppigrf = ppigrf.igrf(*coords, date)
    times_ppigrf.append(time.perf_counter() - start)
median_ppigrf = np.median(times_ppigrf)
print(f"ppigrf          : {median_ppigrf:.3f} +- {np.std(times_ppigrf):.3f} s ")
```

This was as much as my computer could handle in terms of RAM from ppigrf. It used almost 8 Gb to calculate on the ~700k points because of all the numpy broadcasting and pandas operations they do. Harmonica and SHTools used virtually nothing.

harmonica/_spherical_harmonics/igrf.py
```diff
@@ -0,0 +1,696 @@
...
+    doi : str
+        The DOI used to download the coefficient file.
+    file_name : str
+        The name of the coefficient file in the online archive.
+    hash : str
+        The hash of the coefficient file, used to check download integrity.
+    reference_radius : float
+        The reference radius used in the spherical harmonic expansion in
+        meters.
+    coefficients : tuple = (g, h)
```
mdtanker, line 210: It seems like the coefficients always have a shape of 14x14, regardless of what value is passed to `max_degree`.

PR #569, opened 2025-07-03
Add grid size check for corrupted .grd files (#451)

### LINKED ISSUE #451: Raise error when GRD file is corrupted

In case the GRD file that is intended to be read is corrupted and there are missing elements in the grid, our `hm.load_oasis_montaj_grid` function will fail, raising an error complaining about not being able to reshape the grid to the expected shape.

We could improve the user interface by checking if the number of elements in the read 1d array matches the expected shape and raise an error with an informative error message if it doesn't. It could be something along: 

```python
f"Encounter grid with {n} elements that doesn't match the expected shape ({shape}). The file {filename} is possibly corrupted."
```

The idea for this new feature was triggered after a conversation with @marcoutojr

Yes, sure! Although I'd love to see anyone else to tackle this issue.

Nanakjoth:
### Summary

Add a simple check in `load_oasis_montaj_grid` to raise a clear `ValueError` if the number of grid elements does not match the expected shape. This helps users quickly see if a `.grd` file is corrupted or incomplete.

---

### What changed

- Added an explicit element count check before `.reshape()`.
- Raises a clear error instead of NumPy’s reshape error.
- Manually tested by creating a fake `.grd` with missing elements.

Closes #451.

### DIFF
--- a/doc/gallery_src/transformations/reduction_to_pole.py
+++ b/doc/gallery_src/transformations/reduction_to_pole.py
@@ -29,7 +29,11 @@ inclination, declination = -52.98, 6.51
 # that the sources share the same inclination and declination as the
 # geomagnetic field.
 rtp_grid = hm.reduction_to_pole(
-    magnetic_grid, inclination=inclination, declination=declination
+    magnetic_grid,
+    inclination=inclination,
+    declination=declination,
+    magnetization_inclination=inclination,
+    magnetization_declination=declination,
 )
 # Show the reduced to the pole grid
 print("\nReduced to the pole magnetic grid:\n", rtp_grid)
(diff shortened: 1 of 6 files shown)


santisoler:
Hi @Nanakjoth. Thanks for taking the time to work on this and opening the PR!

I like the error message you added. Before we merge this PR I would like to add some test that checks that the error is being raised when a corrupted file has been passed. We could add the sample corrupted file you used to manually test the function to `harmonica/tests/data`, name it something like `incomplete_grid.grd`, and add a test function in `harmonica/tests/test_oasis.py` that tries to load the file and checks your error was raised.

We can use [`pytest.raises`](https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions) to run the check.

Would you like to work on that test?

Nanakjoth:
Hi @santisoler,
Absolutely, I would love to do it.

I actually thought about adding it at the same time as the fix but didn’t get around to it, so thanks for suggesting, that helps a lot.

Will update the PR shortly! 🚀

Issue #594, opened 2025-11-12
Improvements to ellipsoids


santisoler:
**Description of the desired feature:**

Listing here some ideas to improve ellipsoids code:

- [ ] Add a `Sphere` class that represents an homogeneous sphere as a source.
  - Use that class in tests, when comparing ellipsoids to spheres.
- [ ] Expose the rotation matrix through a property in ellipsoid classes.
- [ ] Move the physical properties of the ellipsoids to the objects. Remove them from the functions calls.
- [ ] Add a constructor function (e.g. `get_ellipsoid`) that takes semiaxes lengths and returns and object of the right class (with semiaxes in the right order).

**Are you willing to help implement and maintain this feature?** 

Sure! I'll be continuing the work that @KellySavanna started in #568. Anyone is more than welcomed to help!

PR #597, opened 2025-11-13
Move physical properties to ellipsoid classes

santisoler:
Include physical properties of the ellipsoids as attributes in their classes: `density`, `susceptibility` and `remanent_mag`. Remove the physical property arguments in `ellipsoid_gravity` and `ellipsoid_magnetic`. Add sanity checks for ellipsoids' arguments: positive semiaxes lengths, correct values when updating them through the setters, types and lengths of physical properties. Create a new `harmonica.typing` module and add a new `Ellipsoid` protocol to it. Update tests and improve some docstrings.

Part of #594

### DIFF
--- a/harmonica/_forward/ellipsoids.py
+++ b/harmonica/_forward/ellipsoids.py
@@ -120,6 +120,21 @@ class BaseEllipsoid:
         """
         return get_rotation_matrix(self.yaw, self.pitch, self.roll)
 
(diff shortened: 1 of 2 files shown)


santisoler:
Would be nice to explain better the different options for susceptibility.

santisoler:
Would be also nice to add a check for positive semiaxes lengths.

PR #603, opened 2025-11-14
Optimize ellipsoid's magnetic forward modelling

santisoler:
Vectorize the magnetic forward modelling of a single ellipsoid for many observation points: build the external demagnetization tensors for a single ellipsoid and multiple observation points all at once, harnessing Numpy's vectorized operations. This way we avoid running Python for loops that made the magnetic forward modelling significantly slower.

**Relevant issues/PRs:**

Part of #594

### DIFF
--- a/harmonica/_forward/ellipsoid_magnetics.py
+++ b/harmonica/_forward/ellipsoid_magnetics.py
@@ -475,7 +475,9 @@ def get_demagnetization_tensor_external(
     c: float,
 ) -> npt.NDArray:
     r"""
-    Construct the demagnetization tensor N on external points.
+    Construct the demagnetization tensor(s) N on external observation points.
+
+    Computes multiple demagnetization tensors, one for each external observation point.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


santisoler:
Some small benchmark I ran today:

```python
import time
import numpy as np
import verde as vd
import harmonica as hm


a, b = 50.0, 35.0
center = (0, 0, 0)
susceptibility = 0.2
ellipsoid = hm.ProlateEllipsoid(
    a, b, yaw=0, pitch=0, center=center, susceptibility=susceptibility
)

region = (-200, 200, -200, 200)
shape = (501, 501)
coordinates = vd.grid_coordinates(region=region, shape=shape, extra_coords=a + 10)

external_field = (55_000.0, 71, 12)

n_runs = 10
times = []
for _ in range(n_runs):
    start = time.perf_counter()
    b_field = hm.ellipsoid_magnetic(coordinates, ellipsoid, external_field)
    end = time.perf_counter()
    times.append(end - start)

print(f"Elapsed time: {np.mean(times):.2e} +/ {np.std(times):.2e} s.")
```

Using code in `main` (8c599792326527522499c02064e848c5ebc59264):

```
Elapsed time: 5.42e+00 +/ 2.78e-02 s.
```

Using this branch (e53cabe73ff0bb662b4746cc9315e52fbcc03626):

```
Elapsed time: 1.28e-01 +/ 2.94e-03 s.
```

That's 40X 😎

PR #607, opened 2025-11-19
Update and standardize documentation plots

### LINKED ISSUE #591: Update colorbar for topo-free disturbance
For the topo-free disturbance [example](https://www.fatiando.org/harmonica/v0.7.0/gallery/gravity_disturbance_topofree.html#sphx-glr-gallery-gravity-disturbance-topofree-py), it would be good to change to colorbar to hinge at 0.

Code for example  here: https://github.com/fatiando/harmonica/blob/main/doc/gallery_src/gravity_disturbance_topofree.py

I'm happy to fix this.

mdtanker:
I've gone through all the documentation and fixed some errors, standardized the plots, and switched all remaining matplotlib plots to use PyGMT. 

Maybe it's personal preferences, but I've switched all diverging color scales to use `balance` instead of `vik` (or `polar`). Based on [this site](https://cmap-docs.readthedocs.io/en/latest/catalog/) `balance` is more perceptually uniform, and I think it also just looks better. 

I fixed some plots that used divering colorscales but either weren't centered on 0 or didn't have equal amplitude max and min limits. 

I've updated the colorscale limits to use actual quantiles of the data , instead of just scaling the max and min values by a factor. 

I've tried to standardize titles, the position and labels of colorbars, and removed some erroneous extra quotations in the titles and labels. 

If anyone doesn't like some of these changes, I'm happy to revert them!

**Relevant issues/PRs:**

Fixes #591


santisoler:
Thanks @mdtanker! I'll check these plots out! I'm updating this branch to bring in the changes in #601 here, so we can see the built documentation by downloading from the uploaded artifact.

santisoler (approved):
Thanks @mdtanker for opening this PR!

I couldn't find the statement about the `balanced` colormap being more perceptually uniform in the website you shared. But nonetheless, my eyes see more contrast in those plots, and therefore I found it easier to see (specially for colors close from the white center).

I saw you replaced some matplotlib plots for pygmt plots in the grid transformation page (and in a few other pages). My reasoning was to plot plain coordinates grids with matplotlib and geographic coordinates grids with pygmt. For plain coordinates the two should produce the same plot, and most users are probably more familiar with matplotlib than with pygmt. Do you think we should make the move to all pygmt plots?

I left two minor comments below, let me know what do you think!

mdtanker:
> I couldn't find the statement about the `balanced` colormap being more perceptually uniform in the website you shared. But nonetheless, my eyes see more contrast in those plots, and therefore I found it easier to see (specially for colors close from the white center).

Just comparing the Perceptual lightness charts for [balance](https://cmap-docs.readthedocs.io/en/latest/catalog/diverging/cmocean:balance/) and [vik](https://cmap-docs.readthedocs.io/en/latest/catalog/diverging/crameri:vik/). Probably a small difference but I agree it's easy to pick out differences.

> I saw you replaced some matplotlib plots for pygmt plots in the grid transformation page (and in a few other pages). My reasoning was to plot plain coordinates grids with matplotlib and geographic coordinates grids with pygmt. For plain coordinates the two should produce the same plot, and most users are probably more familiar with matplotlib than with pygmt. Do you think we should make the move to all pygmt plots?

Oh I assumed they just hadn't been switched over to pygmt yet! I saw for example, in the [prism forward modeling user guide](https://www.fatiando.org/harmonica/latest/user_guide/forward_modelling/prism.html) there was already a mix of pygmt and matplotlib. 

I'm happy to revert to matplotlib, but then we would probably switch those pygmt plots to matplotlib as well. 

One thing I like about using pygmt is it shows users how to go from separate coords / data to an xarray dataarray with `vd.make_xaray_grid`. What are your thoughts?

santisoler:
> I'm happy to revert to matplotlib, but then we would probably switch those pygmt plots to matplotlib as well.

I don't have strong opinions on having matplotlib or pygmt plots. Actually, I find myself going back to the Harmonica's pygmt plots when I don't remember how to do certain thing in pygmt. My point is that it's not necessary to switch all to matplotlib or all to pygmt. I do prefer pygmt for data we already have in geographical coordinates, though. And I think it's nice to have some matplotlib plots in the user guide, so readers that are already familiar with matplotlib only need to get used to one single new thing (Harmonica), rather than learning Harmonica _and_ how to use pygmt.

> One thing I like about using pygmt is it shows users how to go from separate coords / data to an xarray dataarray with `vd.make_xaray_grid`. What are your thoughts?

We can use that function in the matplotlib plots too. We can define the dataarray and plot it directly through the [`DataArray.plot.pcolormesh`](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.plot.pcolormesh.html) method. We can use the function and comment that we can optionally convert the grid to a `DataArray` and then plot it (using either matplotlib or pygmt).

mdtanker:
Hi @santisoler this should now be ready for review! I reverted all the projected coordinate examples which were plotted using matplotlib back to how they were originally. I think now there is a decent mix of plotting styles:
- using the built-in xarray `.plot()`
- using `pcolormesh`
- using pyGMT with both geographic and projected coordinates

I have used the new verde percentile parameters in `vd.maxabs` and `vd.minmax`, which required adding a min verde version to the `environment.yml` file. Is this ok? 

Happy to revert anything back to how it was!

The docs can be downloaded here: https://github.com/fatiando/harmonica/actions/runs/25309776972/artifacts/6780042235

environment.yml
```diff
@@ -27,7 +27,6 @@ dependencies:
   - pytest
   - pytest-cov
   - coverage
-  - boule
```
santisoler, line 30: I'd restore `boule` here, since we need it for running some tests.

environment.yml
```diff
@@ -40,6 +39,7 @@ dependencies:
   - pyproj
   - matplotlib
   - ensaio>=0.5.0
+  - verde>=1.9.0
```
santisoler: I'd add a small comment here. I can see myself trying to remove one of the two verde items in the future without remembering why we added it in the first place 😅 

```suggestion
  - verde>=1.9.0  # needed for some utilities used in docs
```

doc/user_guide/topographic_correction.rst
```diff
@@ -260,6 +264,33 @@ And plot it:
    fig.colorbar(cmap=True, frame=["a50f25", "x+lTopography-free gravity disturbance", "y+lmGal"])
    fig.show()
 
+Compare the Bouguer and Topography-free disturbances
+----------------------------------------------------
```
santisoler, line 267: Matt, do you mind adding some explanation of what we are plotting here, and some minor conclusion. Like for example that there's an appreciative difference between using the simple Bouguer plate correction and forward modelling the terrain.

Also, it seems to me that this plot only has positive values. In such case, wouldn't it make more sense to use a non-divergent colorscale (like viridis) to plot such difference. I think it would add more contrast to the plot.

Image

doc/user_guide/equivalent_sources/block-averaged-eqs.rst
```diff
@@ -111,14 +111,14 @@ And project the geographic coordinates to plain Cartesian ones:
             style="c0.1c",
             cmap=True,
         )
-    fig.colorbar(cmap=True, position="JMR", frame=["a200f100", "x+lnT"])
+    fig.colorbar(cmap=True, position="JMR+e", frame=["a200f100", "y+lnT"])
     fig.show()
 
 
-Most airborne surveys like this one present an anysotropic distribution of the
+Most airborne surveys like this one present an anisotropic distribution of the
```
santisoler, line 118: Thanks for catching that!

santisoler (changes requested):
Looking great to me, @mdtanker! I just left a few minor suggestions. Feel free to merge once you address them.

Thanks again for taking the time to do this! 🙏🏼

environment.yml
mdtanker: Ok I can restore, but it is listed above, in the testing section, in case you didn't see it there.

doc/user_guide/topographic_correction.rst
mdtanker: Good call, I've added some more info and updated the plots. Let me know what you think!

environment.yml
santisoler: Ha, true. Sorry, I missed it! Let's keep only one `boule` in the list.

doc/user_guide/topographic_correction.rst
```diff
@@ -244,22 +247,79 @@ And plot it:
...
+Now that we have computed the Bouguer and Topography-free disturbances, we
+can plot the difference to get a sense of how much the methods differ.
+From the below plot, we can see large differences (up to 10 mGal) arise from the
+different methods. Generally, these large differences are in regions of rugged
+topography, while the regions of flat topography have smaller  differences (<2 mGal).
+Additionally, almost all of the differences are positive, meaning the Bouguer
+disturbance is generally smaller than the Topography-free disturbance.
+
+This is because the flat-slab assumption always overestimates the correction. It
+doesn't account for a valleys below, or terrain above, the observation points, both
```
santisoler: Minor improvement to wording here.
```suggestion
doesn't account for valleys below or terrain above the observation points, both
```

doc/user_guide/topographic_correction.rst
```diff
@@ -244,22 +247,79 @@ And plot it:
...
+of which decrease the observed gravity. Not accounting for these results in too
+large of a Bouguer correction, and therefore too small of a Bouguer disturbance.
+To account for this, additional estimated terrain corrections are often included
+in the Bouguer correction. This highlights the benefit of the forward modelling
```
santisoler: I think I'd remove this line to avoid confusion. I've seen people often applying a Bouguer plate correction and then performing a terrain correction on a portion of it. I think this makes the process more complex without any real gain. If the forward for the topography is needed, then why not just model it entirely without adding the extra step of the Bouguer plate.
```suggestion
This highlights the benefit of the forward modelling
```

santisoler (approved):
Looking great @mdtanker! I left two small comments regarding the new paragraphs you added to the terrain correction page. Feel free to accept them and edit them as you will. I think you can merge this afterwards if there's nothing else you want to do here.

PR #608, opened 2025-11-19
Pad grids automatically before applying transformations

### LINKED ISSUE #390: Put padding inside the transformation function

Right now the padding process is done externally before passing it to the grid transformation
padding -> hm.filter -> unpadding 
I'm thinking to put the padding and unpadding inside the transformation, so we can use one-line code to do the filter
where paddling and unpadding are handled inside the transformation.

Probably we can also work on different type of padding method, for example [Fedi, 2012](https://academic.oup.com/gji/article/188/1/103/632206). But I guess this will be better to implement on the [xrft ](https://github.com/xgcm/xrft) side?
What do you guys think?

Happy to implement this, but also love to see contributors to tackle this one

### LINKED ISSUE #630: Extend test to the reduction to the pole transformation

Currently the reduction to the pole filter and transformation are being tested with a re-implementation of the filter in the frequency domain, and a comparison of the results with an RTP grid created with Oasis Montaj.

It would be nice if we can add an extra test that uses a synthetic model (like a single dipole) to produce a grid with arbitrary inclination, declination (both for the inducing field and for the source itself), and a grid with the same source in the pole. Then we can run our RTP function on the first grid and compare it with the latter.

Sure! But I'd love to see anyone else to tackle this!

leouieda:
Rework the `apply_filter` function to make it apply and then remove padding when applying a filter automatically. The padding can be turned off or controlled by passing keyword arguments. The default padding mode is "edge" which works well with our tests and examples. Dropping non-dimensional coordinates is now handled by `apply_filter` as well instead of the FFT functions. Only drop non-dimensional coordinates on upward continuation since that's the only transformation that changes coordinates. Update the tests for transformations, in particular the tilt, Gaussian filters, and reduction to the pole which now test against a synthetic instead of the Montaj results from previously. Remove the unused Montaj testing grid.  

**Relevant issues/PRs:** Fixes #390 and fixes #630

### DIFF
--- a/harmonica/_transformations.py
+++ b/harmonica/_transformations.py
@@ -96,9 +96,9 @@ def derivative_easting(
         evenly spaced (regular grid). Its dimensions should be in the following
         order: *northing*, *easting*. Its coordinates should be defined in the
         same units.
-    order : int
+    order : int, optional
         The order of the derivative. Default to 1.
-    method : str (optional)
+    method : str, optional
         Method that will be used for computing the easting derivative. It can
         be either ``"finite-diff"``, for computing using
         :func:`xarray.differentiate`, or ``"fft"``, for using FFT-based
@@ -172,9 +172,9 @@ def derivative_northing(
         evenly spaced (regular grid). Its dimensions should be in the following
         order: *northing*, *easting*. Its coordinates should be defined in the
         same units.
-    order : int
+    order : int, optional
         The order of the derivative. Default to 1.
-    method : str (optional)
+    method : str, optional
         Method that will be used for computing the easting derivative. It can
         be either ``"finite-diff"``, for computing using
         :func:`xarray.differentiate`, or ``"fft"``, for using FFT-based


harmonica/_transformations.py
```diff
@@ -175,13 +221,18 @@ def derivative_northing(grid, order=1, method="finite-diff"):
     return grid
 
 
-def upward_continuation(grid, height_displacement):
+def upward_continuation(grid, height_displacement, pad=True, pad_kwargs=None):
```
santisoler:
```suggestion
def upward_continuation(grid, height_displacement, *, pad=True, pad_kwargs=None):
```

harmonica/_transformations.py
```diff
@@ -193,6 +244,16 @@ def upward_continuation(grid, height_displacement):
         The height displacement of upward continuation. For upward
         continuation, the height displacement should be positive. Its units
         are the same units of the ``grid`` coordinates.
+    pad : bool
```
santisoler:
```suggestion
    pad : bool, optional
```

harmonica/_transformations.py
```diff
@@ -265,6 +347,16 @@ def gaussian_highpass(grid, wavelength):
     wavelength : float
         The cutoff wavelength in high-pass filter. Its units are the same
         units of the ``grid`` coordinates.
+    pad : bool
```
santisoler:
```suggestion
    pad : bool, optional
```

harmonica/filters/_utils.py
```diff
@@ -29,34 +39,76 @@ def apply_filter(grid, fft_filter, **kwargs):
...
+        ``fft_filter`` in the form of a dictionary.
+    pad : bool
+        If True, will add padding to the grid before taking the Fourier Transform
+        and applying the filter and remove it after the inverse Fourier Transform.
+        Adding padding usually helps reduce edge effects from signal truncation.
+        Default is True.
+    pad_kwargs : dict
+        Any additional keyword arguments that should be passed to the
+        :meth:`xarray.DataArray.pad` function. If none are given, the default
+        padding of 25% the dimensions of the grid will be added using the
```
mdtanker: See comment in review of `_transformations.py`.

```suggestion
        padding of 25% the largest grid dimension will be added using the
```

mdtanker (approved):
These changes look good to me! The only suggestions I have are about the default pad width. The docstrings could be interpreted as the default being 25% of each dimension, meaning a different pad width is used for each dimension, whereas the code uses 25% of the larger of the two dimensions. I added commit suggestions to clarify the docstrings. The other option would be actually to use pad widths per dimension. Since most study regions are typically somewhat near square, this is probably not a big issue, but with a very elongate region, maybe a default of 25% of each dimension's length would be better. I'm happy with either!

mdtanker (approved):
Looks good to me!

PR #611, opened 2025-11-21
Solve numerical instabilities of almost spheroids

santisoler:
Solve the numerical instabilities in gravity and magnetic fields due to prolate, oblate or triaxial ellipsoids that are close enough to a sphere. Fallback to sphere analytic solutions when the ellipsoid's semiaxes lengths are close enough within a relative tolerance.

**Relevant issues/PRs:**

Part of #594


santisoler:
TODO:

- [ ] Add tests that would fail without this patch

PR #617, opened 2025-12-11
Update and refactor code and docs for tesseroids

mdtanker:
This PR fixes and clarifies the documentation about how to use tesseroids and tesseroid layers. It also tries to align the functionality of `prisms` and `tesseroids`, by adding a `_discard_thin_tesseroids()` function.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -126,7 +126,7 @@ jobs:
 
       # Fetch the built docs from the "build" job
       - name: Download HTML documentation artifact
-        uses: actions/download-artifact@v7
+        uses: actions/download-artifact@v8
         with:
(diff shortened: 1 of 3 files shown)


santisoler:
Thanks @mdtanker for revisiting the tesseroid code! Since it involves several files, maybe we could reorganize them similarly to how I organized the [ellipsoids code](https://github.com/fatiando/harmonica/tree/main/harmonica/_forward/ellipsoids). Since the `harmonica._forward` module is private, we can reorganize files as we please. We just need to remember to import the public functions in `harmonica/__init__.py`.

mdtanker:
I think there were a few other things I may have not push yet but probably won't be able to get around to it for a week or 2.

PR #662, opened 2026-04-11
correct a error string and add a physical prerequisite check in isost…

nameearly:
…atic_moho_airy that density_mantle > density_crust


leouieda:
This is an AI agent.

PR #666, opened 2026-04-15
Add citation for EulerInversion in the "citing" page

leouieda:
We list our publications that should be cited when certain methods are used. There was none for the new Euler inversion implementation.


santisoler:
Nice catch, @leouieda. Yes, we should add it to the list so we remind users to cite the paper!

PR #669, opened 2026-04-15
Increase fetch depth on PyPI Action to fix version numbers

leouieda:
When we publish built packages to TestPyPI, the version number is wrong because we passed the 100 commit depth we had established earlier. When we do this, setuptools_scm can't find the last tag and then it can't generate the right version number. That's how we ended up with https://test.pypi.org/project/harmonica/0.0.post100/. Increase the depth so this doesn't happen. I don't think this would affect the version number for an actual release but it's best to be safe.


santisoler:
Nice catch, @leouieda. Better be safe than sorry. I'll merge this.

PR #673, opened 2026-04-20
Use Bordado in code, tests and docs

santisoler:
Add `bordado` as required dependency. Replace some of the Verde functions (like `grid_coordinates`, `get_region`, `rolling_window`, etc.) for their Bordado counterparts in source code, tests, and docs. Update the code that uses those functions: keep in mind that Bordado functions are n-dimensional, so additional coordinates are not ignored as in Verde ones.

**Relevant issues/PRs:**

Fixes #656


doc/user_guide/equivalent_sources/eqs-parameters-estimation.rst
```diff
@@ -175,12 +175,14 @@ And grid the data using the two equivalent sources:
 
 .. jupyter-execute::
 
+   import bordado as bd
+
    # Define grid coordinates
-   region = vd.get_region(coordinates)
-   grid_coords = vd.grid_coordinates(
+   region = bd.get_region((easting, northing))
```
leouieda, line 181: You could also do 

```suggestion
   region = bd.get_region(coordinates[:2])
```

but it's not much simpler. But just in case you don't want to use `(easting, northing)` all the time.

src/harmonica/_equivalent_sources/cartesian.py
```diff
@@ -268,7 +269,7 @@ def _build_points(self, coordinates):
         if self.block_size is not None:
             coordinates = self._block_average_coordinates(coordinates)
         if self.depth == "default":
-            self.depth_ = 4.5 * np.mean(vd.median_distance(coordinates, k_nearest=1))
+            self.depth_ = 4.5 * np.mean(median_distance(coordinates, k_nearest=1))
```
leouieda:
```suggestion
            self.depth_ = 4.5 * np.mean(bd.neighbor_distance_statistics(coordinates[:2], "median", k=1))
```

This function will probably replace the `median_distance`.

src/harmonica/_equivalent_sources/cartesian.py
```diff
@@ -12,11 +12,12 @@
 
 import warnings
 
+import bordado as bd
 import numpy as np
-import verde as vd
 import verde.base as vdb
 from numba import jit
 from sklearn.utils.validation import check_is_fitted
+from verde import BlockReduce, median_distance
```
leouieda: I'd keep the `import verde as vd` to maintain the clear namespace separation.

src/harmonica/_equivalent_sources/gradient_boosted.py
```diff
@@ -14,8 +14,8 @@
 
 import numpy as np
 import verde.base as vdb
+from bordado import get_region, rolling_window
```
leouieda: I'd replace this with `import bordado as bd` to keep the namespace.

src/harmonica/_forward/prism_layer.py
```diff
@@ -11,8 +11,8 @@
 import warnings
 
 import numpy as np
-import verde as vd
 import xarray as xr
+from verde import make_xarray_grid
```
leouieda: Same about the import.

src/harmonica/_spherical_harmonics/igrf.py
```diff
@@ -11,11 +11,12 @@
 import datetime
 import pathlib
 
+import bordado as bd
 import boule
 import numba
 import numpy as np
 import pooch
-import verde as vd
+from verde import make_xarray_grid
```
leouieda: Same here as well.

leouieda (changes requested):
Looks great! Made some small suggestions. I also noticed that Bordado is missing from the list in https://www.fatiando.org/harmonica/dev/install.html#dependencies (Boule is too, actually).

doc/user_guide/equivalent_sources/eqs-parameters-estimation.rst
santisoler: Yes! I think I prefer explicitly passing `easting` and `northing` in the guide. Since we are already defining those variables, it's not a big deal.

src/harmonica/_equivalent_sources/cartesian.py
santisoler: Right! Nice catch!

doc/user_guide/equivalent_sources/eqs-parameters-estimation.rst
leouieda: 👍🏽

leouieda (approved):

PR #675, opened 2026-04-22
Add `Ellipsoid.get_surface` method

santisoler:
Add a method to the `Ellipsoid` class to extract a set of discrete points located on its surface. The main purpose of this method is to be able to plot the ellipsoid in libraries like Matplotlib. Add a test that checks if the surface points satisfy the characteristic equation of ellipsoids.

### DIFF
--- a/src/harmonica/_forward/ellipsoids/ellipsoids.py
+++ b/src/harmonica/_forward/ellipsoids/ellipsoids.py
@@ -326,7 +326,7 @@ class Ellipsoid:
         ellipsoid.translate(self.center, inplace=True)
         return ellipsoid
 
-    def get_surface(self, *, size=181):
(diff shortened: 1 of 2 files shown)


src/harmonica/_forward/ellipsoids/ellipsoids.py
```diff
@@ -325,3 +325,67 @@ def to_pyvista(self, **kwargs):
         ellipsoid.rotate(rotation=self.rotation_matrix, inplace=True)
         ellipsoid.translate(self.center, inplace=True)
         return ellipsoid
+
+    def get_surface(self, *, size=181):
+        """
+        Return three arrays with a discrete representation of the ellipsoid surface.
```
santisoler:
```suggestion
        Return three arrays with a discrete representation of the ellipsoid's surface.
```

Issue #646, opened 2026-03-25
Issue with xrft for next release


mdtanker:
**Description of the problem:**

I'm trying to publish a new package on conda-forge which relies on harmonica and am having issues creating a working recipl.yml due to an issues with the latest release of `xrft`, v1.0.2. I think this is linked to this issue: https://github.com/xgcm/xrft/issues/221. When this version is installed, the system seems to think the installed version is actually 0.0.0 instead of 1.0.2 which then causes compatibility errors since harmonica requires >1.0. Before we make our less release, hopefully the above linked issues is fixed, or if not we may need to pin our dependecy to >1.0,<1.0.2. 

Not sure if this is an actually issue that requires a fix now, or just before the next release. I can  try and create an  new environment and see if I get an issue.

santisoler:
I suspect that since the issue is with the released wheel, they'll have to patch it, release `1.0.3`, and yank `1.0.2` so it never gets installed.

santisoler:
I tried rerunning the build process in the conda-forge feedstock repo and apparently I don't get any error: https://github.com/conda-forge/harmonica-feedstock/pull/16. [This build](https://github.com/conda-forge/harmonica-feedstock/runs/72564634623) was run without any change to the recipe.

@mdtanker, have you check recently if your package is building now? Or are you still facing the same issue with `xrft`?

mdtanker:
Looks to still be failing, with this error: [harmonica 0.7.0 has requirement xrft>=1.0, but you have xrft 0.0.0.](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1510528&view=logs&j=4cabe686-70ae-553a-7fd0-310379f2cbac&t=6a4fc7c9-c31a-5115-eff9-6479d72b69ff&l=1272)

I'm using the new version of the conda-forge recipe (`recipe.yaml` instead of meta.yaml`) and harmonica host requirements have `setuptools` and `setuptools_scm` which I don't have, so maybe it's something to do with that? 

Edit: I just add `setuptools` and still recieved the same error.

leouieda:
I think the issue is that the xrft feedstock is fetching from GitHub instead of PyPI. The version number isn’t present in GitHub and is only generated when a source dist or wheel is built. A solution would be to change the xrft recipe to fetch from PyPI.

leouieda:
[...]

PR #677, opened 2026-04-27
Implement our own FFT functions for `xr.DataArray`

santisoler:
Add public `harmonica.filters.fft` and `harmonica.filters.ifft` functions to compute the FFT and its inverse of a `xr.DataArray`, respectively. Remove `xrft` as a dependency, but keep it as a requirement to run the tests. Add `harmonica.filters.pad` and `harmonica.filters.unpad` functions as modified versions of the ones in `xrft`. Add tests for the new functions: the FFT functions are tested against their `xrft` counterparts, while the tests for the padding functions that were already existing in `xrft` were included. Skip doctests for some padding functions since Xarray changed the order in which the coordinates are displayed when printing a `xr.DataArray` since a few versions ago.

**Relevant issues/PRs:**

Related to #646

TODO:

- [x] Write tests for the new `fft` and `ifft` functions
- [x] Remove `xrft` as a dependency
- [x] Implement the padding functions.
- [x] Replace `xrft` for the new functions throughout the codebase.
- [x] Make the fft functions public
- [x] Add tests. Maybe for sake of time, just compare with xrft results.
  - [ ] Open issue to add our own tests against analytical solutions.


leouieda:
@santisoler could you make the FFT functions public? I've been using xrft a lot in classes along with Harmonica and it would be useful to have this.

src/harmonica/filters/_fft.py
```diff
@@ -22,60 +27,213 @@ def fft(grid, true_phase=True, true_amplitude=True, **kwargs):
...
+
+    # Generate new coordinates
+    fft_dims = tuple(f"{prefix}{dim}" for dim in grid.dims)
+    fft_coords = {
+        f"{prefix}{coord}": (dim, _fftfreq(grid.coords[coord]))
+        for coord, dim in zip(dimensional_coords, fft_dims, strict=True)
+    }
+
+    # Compute FFT
+    fft = np.fft.fftshift(np.fft.fftn(grid.values))
```
leouieda, line 99: @santisoler it would be good to include parts of xrft that calculate the theoretical Fourier Transform instead of just the FFT result. That's in the `true_phase` and `true_amplitude` parts of the code: https://github.com/xgcm/xrft/blob/master/xrft/xrft.py#L462 

For amplitude based filters this doesn't matter too much since it's all relative. But for spectrum analysis and other possible filters it could matter. So maybe we should add it always to be safe.

santisoler:
I think this is quite ready for our immediate goals of removing `xrft` as a dependency.

In the process, I've been thinking about improving how non-dimensional coordinates are handled. I don't quite like the fact that users need to drop coordinates before passing the `xr.DataArray` to the FFT or padding functions.

Maybe we should assume that the padding will only get applied to dimensional coordinates (coordinates that have the same name as the dimension, i.e. they are `Index`), and keep the other coordinates padded with `nan`s (as the default behaviour in `xarray.DataArray.pad`. The `unpad` function could remove additional elements from all coordinates, so a `unpad(pad(grid))` will be truly a roundtrip even if the `grid` contains non-dimensional coordinates.

Similar thing with FFT: we could only compute frequencies using the dimensional coordinates and ignore the other ones when building the grid in the frequency domain.

I'll open an issue for this after merging this PR.

leouieda:
> Maybe we should assume that the padding will only get applied to dimensional coordinates (coordinates that have the same name as the dimension, i.e. they are Index), and keep the other coordinates padded with nans (as the default behaviour in xarray.DataArray.pad. The unpad function could remove additional elements from all coordinates, so a unpad(pad(grid)) will be truly a roundtrip even if the grid contains non-dimensional coordinates.

Yes! 100% agree with this and the FFT round trip with non-dim coords.

src/harmonica/filters/_fft.py
```diff
@@ -22,60 +26,321 @@ def fft(grid, true_phase=True, true_amplitude=True, **kwargs):
...
+    r"""
+    Compute the inverse Fast Fourier Transform of a 2D regular grid.
+
+    If the frequency coordinates have a *shift* attribute, it will be used to shift the
+    coordinates in the spatial domain to such value.
+
+    .. important::
+
+        Assumes that the ``fft_grid`` is *shifted*: it was passed to
+        :func:`numpy.fft.fftshift`. The outputs of the ``fft`` function satisfy this
```
leouieda, line 133:
```suggestion
        :func:`numpy.fft.fftshift`. The outputs of the :func:`harmonica.filters.fft` function satisfy this
```

leouieda:
Just had a minor comment about a docstring. But otherwise, this is good to go! Good work!

PR #679, opened 2026-05-19
Fix cache poisoning vulnerability in GitHub Actions

leouieda:
Zizmor raised an error because of [cache poisoning](https://docs.zizmor.sh/audits/#cache-poisoning), which is where an atacker can hide a malicious payload to cached data and gain access to other workflows when the cache is restored. This could be used to corrupt a release artifact when the release workflow uses caching. We did this in our docs workflow. We also uploaded artifacts of the docs and PyPI distributions and then restored them. Seems like this should also be vulnerable to similar atacks. To avoid all of this, redo the way we deploy in Actions. Add a separate `publish` workflow that only runs on pushes to `main` and releases. It builds the docs and wheels, then publishes them to GitHub pages and PyPI/TestPyPI all in the same job without using caching or artifacts. The `docs` workflow only builds the docs and doesn't deploy them anymore. The `pypi` workflow was removed, with the check of the built packages living in the `test` workflow now.

**Relevant issues/PRs:** Related to https://github.com/fatiando/bordado/pull/115


leouieda:
Updated the PyPI and TestPyPI publishing config to match.

PR #683, opened 2026-06-01
Fix north-south flip in FFT filters for descending-coordinate grids

### LINKED ISSUE #586: harmonica.reduction_to_pole  flipping the output in north-south
Using harmonica.reduction_to_pole (v 0.7.0) with input TMI data read from an ERmapper ERS grid to an xarray. Data is in the north hemisphere in a UTM projection.Plotted input using a matplotlib helper - OK. Pass the xarray to harmonica.reduction_to_pole; plot the output (rtp) with same matplotlib helper). The RTP was computed but the data is flipped north to south. Both data arrays have similar spatial attributes (pasted below). Note that the CRS is dropped in the output. 

[AMF_38N_TestArea_Harmonica.zip](https://github.com/user-attachments/files/22538024/AMF_38N_TestArea_Harmonica.zip)

binggao1230:
Fixes #586

### Problem
The FFT-based filters (`derivative_*`, `upward_continuation`, `reduction_to_pole`, etc., all routed through `apply_filter`) silently produced an output **flipped along a dimension** when the input grid stored that coordinate in **descending** order. This happens for grids read from some file formats (e.g. ERMapper/ERS) where the northing decreases with row index. @santisoler reproduced this in the issue.

### Root cause
In `harmonica/filters/_utils.py::apply_filter`, the inverse FFT (`xrft.ifft`) always returns the grid in **ascending** coordinate order. The code then did:

```python
filtered_grid = filtered_grid.assign_coords(
    {dims[1]: grid[dims[1]], dims[0]: grid[dims[0]]}
)
```

assigning the **original (descending)** coordinates onto the **ascending** result — mislabelling every row/column so the data ends up flipped relative to its coordinates. (The frequency coordinates built by `xrft` also assume positive sample spacing.)

### Fix
Sort the grid to ascending order before transforming, and restore the original coordinate order afterwards via `reindex`. Grids already in ascending order take the same path as before (no `sortby`/`reindex`), so their results are unchanged.

### Reproduction / verification
A minimal synthetic check (no external data needed): apply a filter to a grid and to the same physical grid with one coordinate reversed; after aligning on that coordinate the two results must be identical.

| filter | rel. mismatch (desc vs asc), before → after |
|---|---|
| `derivative_upward` | ~1.0 (flipped) → 0.0 |
| `upward_continuation` | ~1.0 → 0.0 |
| `reduction_to_pole` | ~1.0 → 0.0 |

- Added a parametrised regression test (`northing` and `easting`) in `test/test_transformations.py` that fails on `main` and passes with the fix.
- Full `test/test_transformations.py` + `test/test_filters.py` suite: 62 passed (was 60). `ruff check` / `ruff format --check` clean.

### DIFF
--- a/src/harmonica/filters/_utils.py
+++ b/src/harmonica/filters/_utils.py
@@ -79,6 +79,18 @@ def apply_filter(
     # xrft doesn't know what to do with them.
     non_dim_coords = {c: grid[c] for c in grid.coords if c not in grid.indexes}
     grid = grid.drop_vars(non_dim_coords.keys())
+    # The FFT-based filters assume the grid coordinates are in ascending order:
+    # xrft builds the frequency coordinates from positive sample spacings, and
+    # the inverse transform always returns the grid in ascending order. Grids
+    # read from some file formats store the northing in descending order, which
+    # silently produced a north-south flipped output because the original
+    # (descending) coordinates were assigned to the ascending result (see #586).
+    # Sort the grid to ascending order before transforming and restore the
+    # original coordinate order at the end.
+    original_coords = {dim: grid[dim] for dim in dims}
+    needs_reorder = any(bool(grid[dim][0] > grid[dim][-1]) for dim in dims)
+    if needs_reorder:
+        grid = grid.sortby(list(dims))
     if pad:
         # By default, use a padding width of 25% of each grid dimension.
         # Fedi et al. (2012; doi:10.1111/j.1365-246X.2011.05259.x) suggest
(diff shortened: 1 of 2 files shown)


santisoler:
Hi @gaoflow. Thanks for opening the PR. I'm suspecting this contribution made significant use of generative AI software to produce the code and the PR description text. Could you confirm this is the case? Could you disclose how generative AI software was used in the process of opening this PR?

binggao1230:
Yes — you're right, and thanks for asking directly.

I work with an AI coding agent (Anthropic's Claude). It helps me locate and reproduce the bug, drafts the patch and the tests, and drafted this PR description. It runs under my GitHub account and my direction: I review the diff before it goes out, the tests are run locally, and I'm accountable for what's submitted. The mismatch numbers in the table are from real runs, not written prose.

The description is also too long, which is the other half of what you picked up on. That one's a fair hit — I'll keep them shorter.

If Harmonica or Fatiando would prefer not to accept AI-assisted contributions, or wants the disclosure recorded in the commit message or changelog, say which and I'll comply — including closing this if that's the call. #586 is your issue and the fix is four lines, so nothing is lost if you'd rather write it yourself.

santisoler:
Hi @binggao1230, thanks for providing those details. I'm afraid I'm closing this PR due to the significant use of AI coding agent. We prefer to receive human created code.

PR #684, opened 2026-06-01
Make magnetization direction required in public reduction_to_pole

### LINKED ISSUE #440: Source magnetization direction should not be optional in reduction to the pole
One of my biggest [pet peeves](https://www.merriam-webster.com/dictionary/pet%20peeve) in magnetics is people  using reduction to the pole indiscriminately in large areas where there is clearly remanent magnetization present. Many don't even realize that this is a problem. My theory is that this is because most software don't ask for the magnetization direction (or even allow you to set one) and implicitly assume induced magnetization only, leading to the impression that reduction to the pole only needs the field direction.

**I would argue that we can make this assumption more explicit by making the magnetization inclination and declination required arguments of our functions.** Right now, they default to `None` which ends up meaning "induced magnetization" (this link may or may not be obvious as well). Removing the option will also result in less code since we don't have to check that both inc and dec were given.

binggao1230:
Completes #661 (which fixes #440).

### Problem
#661 made the magnetization inclination/declination **required** in `reduction_to_pole_kernel` and removed the code that filled in defaults when they were `None`. However, the **public** `reduction_to_pole` function in `_transformations.py` was missed — it still declared:

```python
magnetization_inclination=None,
magnetization_declination=None,
```

and passed those straight into the now-strict kernel. So the documented default usage (omitting the magnetization direction, as the docstring still described) crashes deep inside the kernel:

```python
>>> hm.reduction_to_pole(grid, inclination, declination)
TypeError: loop of ufunc does not support argument 0 of type NoneType which has no callable radians method
```

This is exactly the silent "induced-only magnetization" foot-gun that @leouieda wanted to remove in #440 — except now it is a crash with a confusing message rather than a clear, explicit requirement.

### Fix
Complete #661 on the public function:
- Make `magnetization_inclination` / `magnetization_declination` required arguments (remove the `=None` defaults).
- Drop the stale *"If None … will be set equal to the inclination/declination"* docstring text, mirroring the kernel docstring updated in #661.

Omitting the magnetization direction now raises a clear error at call time:
```
TypeError: reduction_to_pole() missing 2 required positional arguments:
'magnetization_inclination' and 'magnetization_declination'
```

### Verification
- All existing `reduction_to_pole` tests and gallery/user-guide examples already pass the magnetization angles explicitly (updated in #661), so they are unaffected.
- Added a regression test (`test_reduction_to_pole_magnetization_required`) asserting the arguments are required; it fails on `main` (the old cryptic `TypeError` message does not name the argument) and passes with this change.
- `test/test_transformations.py` + `test/test_filters.py`: 61 passed. `ruff check` / `ruff format --check` clean.

### DIFF
--- a/src/harmonica/_transformations.py
+++ b/src/harmonica/_transformations.py
@@ -389,8 +389,8 @@ def reduction_to_pole(
     grid,
     inclination,
     declination,
-    magnetization_inclination=None,
-    magnetization_declination=None,
+    magnetization_inclination,
+    magnetization_declination,
     *,
     pad=True,
     pad_kwargs=None,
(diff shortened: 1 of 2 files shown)


santisoler:
Closing this due to significant use of AI agent.

PR #689, opened 2026-07-05
Add required memory functions for non-gradient boosted equivalent sources

mdtanker:
I find I commonly want to check whether I can fit a jacobian in memory for regular (non-gradient boosted) equivalent sources, but have no easy way to check whether this is true, of if I need to switch to gradient boosted. This PR adds equivalent `estimate_required_memory` functions for cartesian and spherical equivalent sources.


src/harmonica/_equivalent_sources/spherical.py
```diff
@@ -123,6 +123,56 @@ def __init__(
...
+        """
+        Estimate the memory required for storing the Jacobian matrix.
+
+        Parameters
+        ----------
+        coordinates : tuple of arrays
+            Arrays with the coordinates of each data point. Should be in the
+            following order: (``easting``, ``northing``, ``upward``, ...).
+            Only ``easting``, ``northing``, and ``upward`` will be used, all
+            subsequent coordinates will be ignored.
```
santisoler, line 136:
```suggestion
            Arrays with the coordinates of each data point. Should be in the
            following order: (``longitude``, ``latitude``, ``radius``, ...).
            Only ``longitude``, ``latitude``, and ``radius`` will be used, all
            subsequent coordinates will be ignored.
```

santisoler (changes requested):
It looks good to me, @mdtanker! Thanks for adding this, I get how helpful it can be.

I left a few small suggestions to the docs, feel free to check them out.

Also, I fixed tests failing in Harmonica, so if you update the branch, your tests should pass now.

Thanks again!

test/test_eq_sources_spherical.py
```diff
@@ -300,3 +300,41 @@ def test_error_ignored_args(coordinates_small, data_small, region):
...
+
+@pytest.mark.parametrize("spacing", [100, 500, 1e3])
+def test_memory_estimation(spacing):
+    """
+    Test the estimate_required_memory class method.
+    """
+    region = (-1e3, 5e3, 2e3, 8e3)
+    coordinates = bd.grid_coordinates(
+        region=region, spacing=spacing, non_dimensional_coords=0
+    )
```
santisoler: This is not a big deal for this test, but I'd update the generated coordinates so they are somewhat more realistic in spherical coordinates. So the region is in degrees, and the non dimensional coord is the radii:

```suggestion
    region = (-20, 20, -20, 20)
    coordinates = bd.grid_coordinates(
        region=region, spacing=spacing, non_dimensional_coords=100.0
    )
```

test/test_eq_sources_spherical.py
```diff
@@ -300,3 +300,41 @@ def test_error_ignored_args(coordinates_small, data_small, region):
...
+def test_memory_estimation_supplied_points():
+    """
+    Test the estimate_required_memory class method with user-supplied points.
+    """
+    region = (-1e3, 5e3, 2e3, 8e3)
+    coordinates = bd.grid_coordinates(
+        region=region, shape=(6, 6), non_dimensional_coords=0
+    )
+    # instead of 1 point beneath each datapoint, give custom points
+    points = bd.grid_coordinates(region=region, shape=(3, 3), non_dimensional_coords=0)
```
santisoler:
```suggestion
    region = (-20, 20, -20, 20)
    coordinates = bd.grid_coordinates(
        region=region, shape=(6, 6), non_dimensional_coords=100.0
    )
    # instead of 1 point beneath each datapoint, give custom points
    points = bd.grid_coordinates(region=region, shape=(3, 3), non_dimensional_coords=50.0)
```

santisoler (approved):
All good, @mdtanker. Just spotted a minor thing in the tests for spherical coordinates. Feel free to merge after addressing those! 🚀

santisoler:
Thanks for merging @mdtanker!

PR #692, opened 2026-07-13
Lazy load optional dependencies

santisoler:
Reduce import time of Harmonica by lazily loading optional dependencies like `pyvista`, `vtk`, and `numba_progress`. Add comments to ignore Ruff complaints about having import statements inside functions.

### DIFF
--- a/src/harmonica/_forward/ellipsoids/ellipsoids.py
+++ b/src/harmonica/_forward/ellipsoids/ellipsoids.py
@@ -16,11 +16,6 @@ import numpy.typing as npt
 
 from ..utils import get_rotation_matrix
 
-try:
-    import pyvista
-except ImportError:  # pragma: no cover
-    pyvista = None
-
 
 class Ellipsoid:
     """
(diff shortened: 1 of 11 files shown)


santisoler:
Results of benchmarking import times with `tuna`.

<details>
<summary>In <code>main</code></summary>
current-main
</details>

<details>
<summary>With this PR</summary>
with-lazy-loading
</details>

mdtanker:
Interesting, I didn't realize how much that slowed down the imports! 

@santisoler, just to check, now if a user doesn't have the optional dependency, instead of an ImportError being raised with a clear message about what package is needed ("Missing optional dependency 'pyvista' required for exporting ellipsoids to PyVista."), they just get a blank ModuleNotFoundError error. 

It's still pretty intuitive what they need to install, but wanted to make sure you realized this change.

santisoler:
I was surprised too! Currently most of the import time is taken in importing Verde. Maybe we could reduce Verde's import times in some way.

> @santisoler, just to check, now if a user doesn't have the optional dependency, instead of an ImportError being raised with a clear message about what package is needed ("Missing optional dependency 'pyvista' required for exporting ellipsoids to PyVista."), they just get a blank ModuleNotFoundError error.

Yes, I'm aware that I removed those custom error messages. I only "left" one for `numba_progress`, since the error will come from a private function importing `ProgressBar`, and that would be too cryptic:

https://github.com/fatiando/harmonica/blob/bf62734aa495bc65819efc88e51b32fde23545c6/src/harmonica/_forward/utils.py#L382-L390

But now that you mention, I think that error message will not prompt out, because I'm catching an `ImportError` in the exeption and not a `ModuleNotFoundError`. So thanks for catching it! I'll patch it in a another PR!

Issue #353, opened 2022-09-28
Include a terrain correction in spherical coordinates in the User Guide


santisoler:
After #316 was merged we now can define tesseroid layers. Would be nice to include an example to the User Guide on how we can use the tesseroid layers to perform terrain corrections on spherical coordinates, similar to the one we already have for Cartesian coordinates: https://www.fatiando.org/harmonica/latest/user_guide/topographic_correction.html

PR #695, opened 2026-07-16
Add User Guide section on tesseroid layers

### LINKED ISSUE #352: Include an example on how to use tesseroid layers in the User Guide
After #316 was merged we now can define tesseroid layers. The docs also include a gallery example that makes use of the new feature. Would be nice to have a section in the User Guide, similar to the one we have for the prism layer: https://www.fatiando.org/harmonica/latest/user_guide/forward_modelling/prism.html#prism-layer

manduinca:
Closes #352

Adds a "Tesseroid layer" section to the tesseroid User Guide page, following the same structure as the prism layer section in `prism.rst`. It documents `tesseroid_layer`, the `DatasetAccessorTesseroidLayer` accessor and its `.gravity()` method, with a self-contained runnable example over a region in South America.

I kept it as a section inside `tesseroid.rst` (with a `_tesseroid_layer` label) rather than a new page, since the prism layer is documented the same way and #352 links to that anchor. No changes to the toctree for the same reason.

The one thing I made sure to highlight in an `.. important::` note is the difference from the prism layer: `surface` and `reference` must be given as geocentric radii, not heights, so the example uses `boule.geocentric_radius` to build them.

I ran the example locally and it produces a sensible `g_z` field. I couldn't build the full HTML docs on my machine (no GMT binary for the PyGMT figure), so I'd appreciate a check that the plot cell renders on CI.


santisoler:
Thanks @manduinca for opening this PR! I'll go through it and leave a review soon!

doc/user_guide/forward_modelling/tesseroid.rst
```diff
@@ -313,6 +313,143 @@ Finally, lets plot it:
...
+Moreover, we can use the :meth:`harmonica.DatasetAccessorTesseroidLayer.gravity` method
+to compute the gravitational field of the tesseroid layer on any set of computation
+points.
+
+.. important::
+
+   Unlike the :func:`harmonica.prism_layer`, the ``surface`` and ``reference`` boundaries
+   of a tesseroid layer must be given as **radii** measured from the center of the Earth,
+   not as heights above a reference level.
+   We can use the :meth:`boule.geocentric_radius` method from :mod:`boule` to
```
santisoler: Minor fix to the reference to Boule's method:
```suggestion
   We can use the :meth:`boule.Ellipsoid.geocentric_radius` method from :mod:`boule` to
```

doc/user_guide/forward_modelling/tesseroid.rst
```diff
@@ -313,6 +313,143 @@ Finally, lets plot it:
...
+We need to define a 2D array with the radii of the uppermost *surface* of the layer. We
+will build a synthetic topography and add it to the reference radii so that the ``surface``
+is also expressed as radii from the center of the Earth:
+
+.. jupyter-execute::
+
+   import numpy as np
+
+   topography = 3e3 * np.sin(longitude * np.pi / 20) * np.cos(latitude * np.pi / 20)
+   surface = reference + topography
```
santisoler, line 389: I think it would be nice to add a plot after this cell with the `surface`, so readers can see the synthetic topography within the area and also its values, so it might stick better that those are radii and not oscillations around the ellipsoid.

santisoler (changes requested):
This looks really good @manduinca! Thanks a lot for taking the effort on writing it!

I left two minor suggestions below. Let me know what do you think.

I see that the style check on GitHub Actions are failing, but this is not your fault. I'm fixing it in #699. So don't worry about that!

manduinca:
Thanks for the review @santisoler! I applied your suggestion for the Boule reference and added a plot of the surface right after that cell. The colorbar ends up showing values around 6.37 million meters, so hopefully it's now much clearer that these are radii and not heights. The change of the ellipsoid radius with latitude is also visible in the plot, which I think helps drive the point home. Happy to tweak it if you had something else in mind.

By the way, I just opened #700 with the spherical terrain correction example for the User Guide (#353), which builds on this section. There's a design question in there about how to handle observation points that fall inside the tesseroids, whenever you get a chance.

santisoler:
Thanks for applying the changes @manduinca. I just took the liberty to push some small changes to the guide: I shifted the topography to positive values only, so we don't have tesseroids below the reference. In some situations, we usually assign negative density contrast to tesseroids that fall below the reference (when modelling gravity anomalies), but I didn't want to get into that for this particular example. But I also wanted the anomaly to look with a bit more variation.

I also added a plot of the topography next to the one for the surface, so readers can see the difference between them.

Let me know what do you think. After this I think this is ready to be merged.

manduinca:
Thanks @santisoler! The changes look great. Shifting the topography to positive values keeps the example simple (no need to get into negative density contrasts here), and plotting the topography next to the surface makes the difference between the two much clearer. From my side this is ready to be merged.

santisoler (approved):
Thanks a lot @manduinca for addressing my comments. I'm merging this one (even the code-style fails due to the issue I'll fix in #699) 🚀 

Congrats for the hard work, and sorry for my usual delays on replying!

PR #698, opened 2026-07-24
Add linearity tests for forward models

### LINKED ISSUE #293: Add linearity tests for forward models

Our forward models for the different geometries (points, prisms, tesseroids) allow to take multiple bodies as inputs, returning a single array as the resulting field generated by every one of these bodies on the computation points.
One of the properties of these forward models is their linearity: the resulting field for all the bodies should be equal to the sum of the individual fields.
Although this property is easily readable from the code, it might be wise to add functions that actually test them.
We could add `test_linearity` functions for each geometry that compares the resulting field of passing a set of bodies to the resulting field of computing the fields of each one separately and then sum them up.

Through `pytests.mark.parametrize` we could write a single function per geometry, adding the `field` as a parameter.
For example:

```python
@pytest.mark.parametrize("field", ("potential", "g_z", ...))
def test_linearity(field):
    """
    Test linearity of the forward model
    """
    ...
```

**Are you willing to help implement and maintain this feature?** Yes, but I would love to help anyone motivated to tackle it. 💪🏼

manduinca:
Fixes #293

Adds `test/test_linearity.py` with linearity tests for each geometry (points, prisms, tesseroids), parametrized by field, following the discussion in the issue: the field of a set of sources is compared against the sum of the fields of each source computed separately.

Includes both gravity and magnetic forward models: `point_gravity` (Cartesian and spherical coordinates), `dipole_magnetic`, `prism_gravity`, `prism_magnetic` and `tesseroid_gravity`, using mixed-sign masses/densities/magnetizations.

The newer `ellipsoid_gravity`/`ellipsoid_magnetic` forwards could get the same treatment in a follow-up if you'd like.


test/test_linearity.py
```diff
@@ -0,0 +1,372 @@
...
+            sample_coordinates, sample_prisms, sample_densities, field
+        )
+        expected = sum(
+            prism_gravity(
+                sample_coordinates,
+                sample_prisms[i : i + 1],
+                sample_densities[i : i + 1],
+                field,
+            )
+            for i in range(sample_densities.size)
```
santisoler: Minor suggestion here: we don't need to index the arrays, we can iterate over them with a `zip` function:


```suggestion
                prism,
                density,
                field,
            )
            for prism, density in zip(sample_prisms, sample_densities, strict=True)
```

test/test_linearity.py
```diff
@@ -0,0 +1,372 @@
...
+            sample_coordinates, sample_tesseroids, sample_densities, field=field
+        )
+        expected = sum(
+            tesseroid_gravity(
+                sample_coordinates,
+                sample_tesseroids[i : i + 1],
+                sample_densities[i : i + 1],
+                field=field,
+            )
+            for i in range(sample_densities.size)
```
santisoler: Similarly here:


```suggestion
                tesseroid,
                density,
                field=field,
            )
            for tesseroid, density in zip(sample_tesseroids, sample_densities, strict=True)
```

test/test_linearity.py
```diff
@@ -0,0 +1,372 @@
...
+    def test_linearity(self, sample_coordinates, sample_points, sample_masses, field):
+        """
+        Compare the field of all points vs. the sum of their individual fields.
+        """
+        result = point_gravity(sample_coordinates, sample_points, sample_masses, field)
+        expected = sum(
+            point_gravity(
+                sample_coordinates,
+                tuple(p[i : i + 1] for p in sample_points),
+                sample_masses[i : i + 1],
```
santisoler: I would suggest indexing the arrays directly, so we pass only a float for each coordinate and for the mass. This way we are also testing (although we are probably already testing that in other places) that passing just a float as coordinates and single float for mass also works.

```suggestion
                tuple(p[i] for p in sample_points),
                sample_masses[i],
```

test/test_linearity.py
```diff
@@ -0,0 +1,372 @@
...
+            region=(-100, 100, -70, 70), spacing=20, non_dimensional_coords=radius
+        )
+        result = point_gravity(
+            coordinates, points, sample_masses, field, coordinate_system="spherical"
+        )
+        expected = sum(
+            point_gravity(
+                coordinates,
+                tuple(p[i : i + 1] for p in points),
+                sample_masses[i : i + 1],
```
santisoler: Similarly here:
```suggestion
                tuple(p[i] for p in points),
                sample_masses[i],
```

test/test_linearity.py
```diff
@@ -0,0 +1,372 @@
...
+        result = dipole_magnetic(
+            sample_coordinates, sample_dipoles, sample_magnetic_moments, field=field
+        )
+        n_dipoles = sample_dipoles[0].size
+        expected = sum(
+            np.asarray(
+                dipole_magnetic(
+                    sample_coordinates,
+                    tuple(d[i : i + 1] for d in sample_dipoles),
+                    tuple(m[i : i + 1] for m in sample_magnetic_moments),
```
santisoler: And here:
```suggestion
                    tuple(d[i] for d in sample_dipoles),
                    tuple(m[i] for m in sample_magnetic_moments),
```

test/test_linearity.py
```diff
@@ -0,0 +1,372 @@
...
+        """
+        result = prism_magnetic(
+            sample_coordinates, sample_prisms, sample_magnetizations, field=field
+        )
+        expected = sum(
+            np.asarray(
+                prism_magnetic(
+                    sample_coordinates,
+                    sample_prisms[i : i + 1],
+                    tuple(m[i : i + 1] for m in sample_magnetizations),
```
santisoler:
```suggestion
                    sample_prisms[i],
                    tuple(m[i] for m in sample_magnetizations),
```

santisoler (approved):
Hi @manduinca! Sorry for the delayed review. I was busy at a conference and catching up with stuff.

This looks really nice, thanks again for taking the effort and work on it! I left some minor suggestions below. Nothing major, just to make a few portions of the tests more Pythonic. Please, let me know what do you think, and feel free to apply and modify the suggestions as you like.

manduinca:
Thanks for the review @santisoler! I applied all six suggestions: `zip` with `strict=True` in the prism and tesseroid tests, and scalars instead of one-element slices in `point_gravity` (cartesian and spherical), `dipole_magnetic` and `prism_magnetic`. Good point that this way we also cover passing floats for the coordinates and the mass.

Ran `test/test_linearity.py` locally: 32 passed. `ruff check` and `ruff format` come back clean.

manduinca:
@santisoler this one has been approved since 18 August, and after the rebase on 14 September the CI is fully green (14/14) — the earlier red `build` and `style` were from `main` and got fixed in #699.

Is there anything left on my side, or is it just waiting for a merge window?

PR #699, opened 2026-07-29
Support timezone information in spherical harmonic functions

santisoler:
Extend the support of the IGRF classes to dates with timezone information. This fix addresses Ruff's DTZ rules by specifying the timezone when calling `datetime` functions. Use the timezone information in the passed date when defining the start of the epoch: this allows to compute time differences between it and the passed date. Add ignore comments where the `tzinfo` is not relevant. Add some tests to the IGRF functions that cover dates with and without timezone information.

Leaving link to those rules for reference: https://docs.astral.sh/ruff/rules/#flake8-datetimez-dtz

### DIFF
--- a/src/harmonica/_spherical_harmonics/igrf.py
+++ b/src/harmonica/_spherical_harmonics/igrf.py
@@ -120,13 +120,17 @@ def interpolate_coefficients(date, max_degree, years, g, h, g_sv, h_sv):
         raise ValueError(message)
     # Find out which epoch is the last compatible one.
     index = int((date.year - years[0]) // 5)
-    seconds_since_epoch = (
-        date - datetime.datetime(year=years[index], month=1, day=1, tzinfo=date.tzinfo)
-    ).total_seconds()
-    utc = datetime.timezone.utc
+    start_of_the_year = datetime.datetime(
+        year=years[index],
+        month=1,
+        day=1,
+        # Specify start of the year in UTC time if date has timezone information
+        tzinfo=datetime.timezone.utc if date.tzinfo is not None else None,
+    )
+    seconds_since_epoch = (date - start_of_the_year).total_seconds()
     epoch = (
-        datetime.datetime(year=years[index] + 5, month=1, day=1, tzinfo=utc)
-        - datetime.datetime(year=years[index], month=1, day=1, tzinfo=utc)
+        datetime.datetime(year=years[index] + 5, month=1, day=1)  # ruff: ignore[DTZ001]
+        - datetime.datetime(year=years[index], month=1, day=1)  # ruff: ignore[DTZ001]
     ).total_seconds()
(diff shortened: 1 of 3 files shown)


santisoler:
@leouieda would you mind double checking these changes? Particularly the change to `interpolate_coefficients` and the test I added. The good thing is that now the IGRF supports dates with timezone information!

src/harmonica/_spherical_harmonics/igrf.py
```diff
@@ -120,12 +120,17 @@ def interpolate_coefficients(date, max_degree, years, g, h, g_sv, h_sv):
...
-        date - datetime.datetime(year=years[index], month=1, day=1)
-    ).total_seconds()
+    start_of_the_year = datetime.datetime(
+        year=years[index],
+        month=1,
+        day=1,
+        # Specify start of the year in UTC time if date has timezone information
+        tzinfo=datetime.timezone.utc if date.tzinfo is not None else None,
+    )
+    seconds_since_epoch = (date - start_of_the_year).total_seconds()
```
leouieda, line 130:
```suggestion
    start_of_epoch = datetime.datetime(
        year=years[index],
        month=1,
        day=1,
        # Specify start of the year in UTC time if date has timezone information
        tzinfo=datetime.timezone.utc if date.tzinfo is not None else None,
    )
    seconds_since_epoch = (date - start_of_epoch).total_seconds()
```

src/harmonica/_spherical_harmonics/igrf.py
```diff
@@ -120,12 +120,17 @@ def interpolate_coefficients(date, max_degree, years, g, h, g_sv, h_sv):
...
     epoch = (
-        datetime.datetime(year=years[index] + 5, month=1, day=1)
-        - datetime.datetime(year=years[index], month=1, day=1)
+        datetime.datetime(year=years[index] + 5, month=1, day=1)  # ruff: ignore[DTZ001]
```
santisoler, line 132: Try to pass `tzinfo=None` and see if Ruff will complain about it, so we could probably get rid of the ignore comment.

leouieda (approved):

src/harmonica/_spherical_harmonics/igrf.py
santisoler: Ruff DTZ001 complains even if `tzinfo=None`. It complains if a `None` object is passed to it and not a `datetime.timezone` one. I could bypass the check if I do something like:
```python
    tzinfo=None
    epoch = (
        datetime.datetime(year=years[index] + 5, month=1, day=1, tzinfo=tzinfo)
        - datetime.datetime(year=years[index], month=1, day=1, tzinfo=tzinfo)
    ).total_seconds()
```

But I would prefer keeping the `# ruff: ignore` comment instead.

